PR #11, opened 2023-09-29
Tango backend attempt

marcelldls:
## Fastcs tango backend
- I have made forks of Martin’s demo and fastcs respectively 
  - https://github.com/marcelldls/FastCS
  - https://github.com/marcelldls/demo-fast-cs
- Gary has given an initial review and I have included his feedback

You can run the demo by:
- Checking out the tango branches of fastcs and the demo
- Spinning up a tango control system 
  - `yum install podman-compose`
  - `podman-compose up` from the demo directory
- From the dev container
  - Start the tickit device via the vscode debug launch
  - Start tango device server by running `test-demo` from the command line

You can interact with the device in the following manners:
- Via an interactive python terminal:
  - Run `itango3` from the dev container
  - Create device proxy: `dev = DeviceProxy("my/device/name")`
  - then you can use auto complete on dev (e.g. `dev.get_attribute_list()` or `dev.RampRate`)
- Via Jive GUI database administation application
  - `podman pull andygotz/tango-jive:7.19` 
  - `podman run -ti --rm -e DISPLAY=$DISPLAY -e TANGO_HOST=localhost:10000 -v /tmp/.X11-unix:/tmp/.X11-unix --security-opt label=type:container_runtime_t --net=host andygotz/tango-jive:7.19`
  - You may first need to run `xhost +local:podman` to get a window
  - Device interaction via the built in ATKPanel GUI can then be done by right clicking on the device (via the tree) in Jive and selecting `monitor device`
- Alternatively via a Taurus gui I scraped together (`python taurus_gui.py`  from outside dev container) (Its not programmatic, I just made it out of interest to get something similar to the epics gui – the observatory science team claim Taurus and Taranta are the popular Tango gui frameworks)

### DIFF
--- a/src/fastcs/backends/tango/dsr.py
+++ b/src/fastcs/backends/tango/dsr.py
@@ -29,6 +29,8 @@ def _wrap_updater_fget(
     attr_name: str, attribute: AttrR, controller: BaseController
 ) -> Callable[[Any], Any]:
     async def fget(tango_device: Device):
+        assert attribute.updater is not None
+
         await attribute.updater.update(controller, attribute)
         tango_device.info_stream(f"called fget method: {attr_name}")
         return attribute.get()
@@ -56,6 +58,8 @@ def _wrap_updater_fset(
     attr_name: str, attribute: AttrW, controller: BaseController
 ) -> Callable[[Any, Any], Any]:
     async def fset(tango_device: Device, val):
+        assert attribute.sender is not None
+
         await attribute.sender.put(controller, attribute, val)
         tango_device.info_stream(f"called fset method: {attr_name}")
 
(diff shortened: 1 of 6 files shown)


src/fastcs/backends/tango/dsr.py
```diff
@@ -0,0 +1,222 @@
...
+    dsr_instance: str = "MY_SERVER_INSTANCE"
+    debug: bool = False
+
+
+def _wrap_updater_fget(
+    attr_name: str, attribute: AttrR, controller: BaseController
+) -> Callable[[Any], Any]:
+    async def fget(self):
+        await attribute.updater.update(controller, attribute)
+        self.info_stream(f"called fget method: {attr_name}")
```
GDYendell, line 32: I don't see where `info_stream` is defined. What is `self` in this context?

I was just trying to add some very simple tests to add some confidence that I don't break the tango backend going forward, but I am not sure what to pass to this function.

GDYendell: `self` is a `Device` in this context. I have added some type hints and change the parameter name to show this. However these methods are defined dynamically, so static type hints do not work.

GDYendell (approved):

PR #18, opened 2023-11-28
Add HTTPConnection

GDYendell:


GDYendell:
It has become clear that needs to be eiger-specific. It could possibly be refactored into a shared base class when we have a second use case.

PR #22, opened 2024-03-06
Update minimum Python version to 3.11

OCopping:
This will allow for the use of nice Asyncio additions


GDYendell:
Could you also add 3.12 to the CI matrix?

GDYendell:
3.12 didn't work because of softioc dependencies.

PR #23, opened 2024-03-06
Fix pre-commit linting check fail

OCopping:


GDYendell (approved):
I can confirm this fails with latest dependencies on main. I guess a rule must have changed.

PR #34, opened 2024-05-13
Add serial connection

marcelldls:
I tested this against both a virtual serial device (based on https://stackoverflow.com/questions/52187/virtual-serial-port-for-linux) as well as some hardware (Thorlabs MFF usb serial device). To test virtually:
#### Create virtual serial port in terminal 1
 `socat -d -d pty,raw,echo=0 pty,raw,echo=0`

Returns:
```
2024/05/13 09:06:37 socat[14186] N PTY is /dev/pts/9
2024/05/13 09:06:37 socat[14186] N PTY is /dev/pts/10
2024/05/13 09:06:37 socat[14186] N starting data transfer loop with FDs [5,5] and [7,7]
```

#### Connect controller in terminal 2
`test-demo  # port="/dev/pts/10"` - https://github.com/DiamondLightSource/demo-fast-cs/pull/11

#### Send serial device response in terminal 3
Test the position AttrR (can inspect via Phoebus screen)
`echo -e -n "\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01" > /dev/pts/9`
or `echo -e -n "\x02\x02\x02\x02\x02\x02\x02\x02\x02\x02\x02\x02" > /dev/pts/9`

#### Monitor port in terminal 4 (optional)
`cat -v /dev/pts/9`

Can see what the controller sends


src/fastcs/connections/serial_connection.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+class NotOpenedError(Exception):
+    pass
+
+
+@dataclass
+class SerialConnectionSettings:
+    port: str
+    baud: int = 115200
```
GDYendell, line 14: It would be nice if this was an enum of valid baud rates. Even better if it could be imported / built from the serial builtin library so we don't need to maintain it.

src/fastcs/connections/serial_connection.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+class SerialConnection:
+    def __init__(self):
+        self.aioserial_instance = None
+        self._lock = asyncio.Lock()
+
+    async def connect(self, settings: SerialConnectionSettings) -> None:
+        self.aioserial_instance = aioserial.AioSerial(
```
GDYendell, line 23: I see this is the name used in the aioserial examples, but it isn't very clear. Could we use something like stream, port, socket, device?

GDYendell (commented):
I think this looks good. Just a couple of minor suggestions.

pyproject.toml
```diff
@@ -10,12 +10,7 @@ classifiers = [
     "Programming Language :: Python :: 3.11",
 ]
 description = "Control system agnostic framework for building Device support in Python that will work for both EPICS and Tango"
-dependencies = [
-    "numpy",
-    "pydantic",
-    "pvi~=0.8.1",
-    "softioc",
-]
+dependencies = ["aioserial", "numpy", "pydantic", "pvi~=0.8.1", "softioc"]
```
GDYendell, line 13: Could you keep this list expanded? It keeps the git history for each dependency separate.

src/fastcs/connections/serial_connection.py
GDYendell: Agreed that it is too difficult to provide intellisense for the possible baud rates without completly restricting the values, which we don't want to do.

GDYendell:
This is good now, but could you squash everything into one commit? Changing the dependency list and then changing it back again is still losing the history. I think the easiest way to do this (besides lazygit) is

```
git reset --soft ed95aec
git commit --amend --no-edit
```

marcelldls:
> This is good now, but could you squash everything into one commit? Changing the dependency list and then changing it back again is still losing the history. I think the easiest way to do this (besides lazygit) is
> 
> ```
> git reset --soft ed95aec
> git commit --amend --no-edit
> ```

Im happy to do this - Although I would suggest the easiest way is "Squash and merge"

GDYendell:
With squash and merge you don't get a merge commit.

marcelldls:
As you wish. BTW the easy way IMO is `git rebase main -i`

GDYendell (approved):

PR #46, opened 2024-07-24
Add allowed values for creating combo box widgets

GDYendell:


GDYendell:
@jsouter I think this is all that is required to display combo box widgets for enums. It doesn't use mbbi/o, but the effect in the UI is the same and we don't have to worry about converting ints. What do you think?

#40 seems to do more that this by supporting specific EPICS fields. This may require a bit more thought because I don't think we should use the EPICS record fields in the core code. Maybe we should some up with some more descriptive names that map on to EPICS fields and then we can use them for Tango too.

jsouter:
> @jsouter I think this is all that is required to display combo box widgets for enums. It doesn't use mbbi/o, but the effect in the UI is the same and we don't have to worry about converting ints. What do you think?
> 
> #40 seems to do more that this by supporting specific EPICS fields. This may require a bit more thought because I don't think we should use the EPICS record fields in the core code. Maybe we should some up with some more descriptive names that map on to EPICS fields and then we can use them for Tango too.

Yeah I think we should figure out how to handle that, though this seems to work okay for string enums in my testing. As far as I know there's no way to have the dropdowns have a label that is different from the value that it sends to the record, (e.g. a string label on a dropdown that actually sends a integer value to the device), so finding some nice interface to assign ZRVL, ONVL, TWVL etc seems like a natural way to implement EPICS integer enums.

GDYendell:
> As far as I know there's no way to have the dropdowns have a label that is different from the value that it sends to the record, (e.g. a string label on a dropdown that actually sends a integer value to the device)

Ah... Yeah I think this is a limitation in pvi currently. I am not sure we have added support for any device that requires this yet. Did you have a specific case that you started #40 for?

jsouter:
> > As far as I know there's no way to have the dropdowns have a label that is different from the value that it sends to the record, (e.g. a string label on a dropdown that actually sends a integer value to the device)
> 
> Ah... Yeah I think this is a limitation in pvi currently. I am not sure we have added support for any device that requires this yet. Did you have a specific case that you started #40 for?

It came up when trying to make odin-fastcs screens for the detectors (tested with eiger and merlin), which generally expect integer values for the enums (e.g. trigger mode for eiger has the allowed options [0, 1, 2, 3] which get mapped to ['ints', 'inte', 'exts', 'exte'] when passed to/from the detector by the odin adapter. Obviously there is another eiger detector interface for fastcs, but this pops up in a few places in all the odin detectors. Also in the current implementation I need to pass extra metadata from odin to provide a mapping between the allowed values and their string names otherwise all the dropdowns entries will just be non descriptive integers, but that's more of an odin discussion).

GDYendell:
OK that makes sense. This works pretty nicely for devices with enums that accept the descriptive name (eiger), so I am going to merge this for now and in future we can look at how to add descriptive names on top for devices that only accept the integer values.

For odin we should be able to accept the descriptive names everywhere, but currently it has been made to work with EPICS mbb records when we won't need to any more.

PR #54, opened 2024-08-21
Add info tags to define PVI structure from controllers

GDYendell:
Refactor helper functions to preserve PV prefix until later so that PVI PV can use it.
Remove call to SetDeviceName and construct full PVs when creating records.
Add call to create info tags on all records as they are created to add them to the PVI of the controller.
Add call to add sub controller PVI to parent PVI


src/fastcs/backends/epics/ioc.py
```diff
@@ -40,95 +41,174 @@ def run(
...
+        access_mode: Access mode of parameter
+    """
+    record.add_info(
+        "Q:group",
+        {
+            f"{prefix}:PVI": {
+                f"pvi.{name}.{access_mode}": {
+                    "+channel": "NAME",
+                    "+type": "plain",
+                    "+trigger": f"pvi.{name}.{access_mode}",
```
coretl, line 179: We would like to do this long term instead:
https://github.com/PandABlocks/PandABlocks-ioc/issues/120

GDYendell: I think this is done now. I also added a system test to try introspecting the PVI structure - I can add some more specific checks here if you can suggest any.

coretl (approved):

PR #56, opened 2024-09-06
Disable EPICS attributes and commands with too-long PV names

jsouter:
Appears to work now, have included some tests.
Added an enabled flag to the Attribute class, so that when EpicsIOC is called -> calls _create_and_link_attribute_pvs -> attributes can get set to `enabled=False`. I initially tried to delete the attribute and remove it from the mapping (see first commit) but this approach seems simpler. It's not sufficient to prevent the creation of the PV, as the epics gui.py calls _get_single_mapping which still creates the gui control component if we do not check for `enabled` or remove it from the SingleMapping.

### DIFF
--- a/src/fastcs/backends/epics/ioc.py
+++ b/src/fastcs/backends/epics/ioc.py
@@ -128,13 +128,12 @@ def _create_and_link_attribute_pvs(pv_prefix: str, mapping: Mapping) -> None:
             match attribute:
                 case AttrRW():
                     if full_pv_name_length > (EPICS_MAX_NAME_LENGTH - 4):
-                        if full_pv_name_length <= EPICS_MAX_NAME_LENGTH:
-                            print(
-                                f"Not creating PVs for {attr_name} as _RBV PV"
-                                f" name would exceed {EPICS_MAX_NAME_LENGTH}"
-                                " characters"
-                            )
-                            attribute.enabled = False
+                        print(
+                            f"Not creating PVs for {attr_name} as _RBV PV"
+                            f" name would exceed {EPICS_MAX_NAME_LENGTH}"
+                            " characters"
+                        )
+                        attribute.enabled = False
                     else:
                         _create_and_link_read_pv(
                             _pv_prefix, f"{pv_name}_RBV", attr_name, attribute
--- a/src/fastcs/backends/epics/ioc.py
+++ b/src/fastcs/backends/epics/ioc.py
@@ -13,6 +13,8 @@ from fastcs.datatypes import Bool, DataType, Float, Int, String
 from fastcs.exceptions import FastCSException
 from fastcs.mapping import Mapping
 
+EPICS_MAX_NAME_LENGTH = 60
+
 
 @dataclass
 class EpicsIOCOptions:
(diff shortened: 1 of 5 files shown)


jsouter:
I have applied the change to add an enabled flag to the attributes, seems simpler this way. _get_single_mapping has to now check if attributes before it adds them to the SingleMapping.

src/fastcs/backends/epics/ioc.py
```diff
@@ -112,13 +114,34 @@ def _create_and_link_attribute_pvs(pv_prefix: str, mapping: Mapping) -> None:
...
+            full_pv_name_length = len(f"{_pv_prefix}:{pv_name}")
+
+            if full_pv_name_length > EPICS_MAX_NAME_LENGTH:
+                attribute.enabled = False
+                print(
+                    f"Not creating PV for {attr_name} for controller"
+                    f" {single_mapping.controller.path} as full name would exceed"
+                    f" {EPICS_MAX_NAME_LENGTH} characters"
+                )
+                continue
```
GDYendell, line 126: I am not sure why we need the enabled flag. It just needs to no call `_create_and_link_*_pv`, right? What does the enabled flag add?

jsouter: We need some mechanism to tell EpicsGUI.create_gui() not to create controls for PVs that don't get created. There are other options like setting the controller attribute to None (or using delattr, but this didn't work well with pytest), anything that prevents the attribute from being provided by _get_single_mapping.

src/fastcs/backends/epics/ioc.py
```diff
@@ -112,13 +114,34 @@ def _create_and_link_attribute_pvs(pv_prefix: str, mapping: Mapping) -> None:
...
 
             match attribute:
                 case AttrRW():
-                    _create_and_link_read_pv(
-                        _pv_prefix, f"{pv_name}_RBV", attr_name, attribute
-                    )
-                    _create_and_link_write_pv(_pv_prefix, pv_name, attr_name, attribute)
+                    if full_pv_name_length > (EPICS_MAX_NAME_LENGTH - 4):
+                        if full_pv_name_length <= EPICS_MAX_NAME_LENGTH:
```
GDYendell, line 131: I don't think the else of this case is reachable, because of the check [here](https://github.com/DiamondLightSource/FastCS/pull/56/files#diff-5cf291216194a03497811c649effa3cf60205d4587ec541263ec5a1f4d3a5fbeR119), so it could be taken out.

src/fastcs/mapping.py
```diff
@@ -46,15 +46,17 @@ def _get_single_mapping(controller: BaseController) -> SingleMapping:
...
+                        case Put():
+                            put_methods[attr_name] = fastcs_method
+                        case Scan():
+                            scan_methods[attr_name] = fastcs_method
+                        case Command():
+                            command_methods[attr_name] = fastcs_method
             case Attribute():
-                attributes[attr_name] = attr
+                if attr.enabled:
+                    attributes[attr_name] = attr
```
GDYendell, line 59: This is getting quite indented now. It could possibly be simplified with something like:
```suggestion
        match attr:
            case WrappedMethod(fastcs_method=Put(enabled=True) as put_method):
                put_methods[attr_name] = put_method
            case WrappedMethod(fastcs_method=Scan(enabled=True) as scan_method):
                    scan_methods[attr_name] = scan_method
            case WrappedMethod(fastcs_method=Command(enabled=True) as command_method):
                    command_methods[attr_name] = command_method
            case Attribute(enabled=True):
                attributes[attr_name] = attr
```

src/fastcs/backends/epics/ioc.py
GDYendell: Ah, for the GUI OK. Yeah, I am not sure it makes sense that the PV and widget creation are separated like this because they need to match. I think this is fine. I would like to revisit the overall logic of this at some point.

GDYendell (changes requested):
Looks good, just a couple of comments.

src/fastcs/backends/epics/ioc.py
jsouter: I think the logic is correct, though maybe a little ugly. It should only print the warning if we have a PV of length 56 < x <= 60, which would be sufficiently short if we didn't append "_RBV". "continue" gets called above for PVs of length x > 60
The else case should be reached when we have a length x <= 56.

The new test in test_ioc.py tests both these cases I believe.

GDYendell: Sorry, I mean the else of `if full_pv_name_length <= EPICS_MAX_NAME_LENGTH` not `if full_pv_name_length > (EPICS_MAX_NAME_LENGTH - 4)`.

Above you are checking `full_pv_name_length > EPICS_MAX_NAME_LENGTH` and continuing if true, so  if it gets past that `full_pv_name_length <= EPICS_MAX_NAME_LENGTH` will always be true here.

jsouter: Ahh you're right, not sure how I missed that.

GDYendell: Cool. Could you update the tests to cover those two missing lines in src/fastcs/attributes.py and then I think we're good to merge.

jsouter: Assuming you meant to cover the case where Commands with long names get disabled, have pushed that

GDYendell (approved):

PR #66, opened 2024-10-28
Remove tango polling

### LINKED ISSUE #50: Tango does not work with Backend event loop
There are errors about running tasks on the wrong event loop now that the Backend starts the loop and passes it to the specific backend. Tango seems to start its own event loop with no option to pass one in. 

I have tested this against the Eiger driver, which needs to do async initialisation before the backend starts. The tango backend was only previously tested against the demo temp controller, so it is possible this has always been a problem.

## Acceptance Criteria
- Tango Backend runs with no errors

marcelldls:
fixes #50 

At the time of implementation the attribute polling was deferred to the device server and therefore it had its own event loop


marcelldls:
Also I have removed "register_dev" from being on start up called. Device registration logic is probably better left to the library consumer

GDYendell:
This looks good. Just to make sure I understand, is the key thing here removing `polling_period=` from the tango attribute creation? Does that stop tango from creating poll tasks? What does adding `green_mode=` do if tango is not doing any polling?

The tests look quite neat, but they are also a bit hard to understand. Could you add some comments / docstrings to the test utilities to explain what they are doing and maybe link to the pytango docs to explain the magic `DeviceTestContext` is doing?

marcelldls:
> This looks good. Just to make sure I understand, is the key thing here removing `polling_period=` from the tango attribute creation? Does that stop tango from creating poll tasks? What does adding `green_mode=` do if tango is not doing any polling?
> 
> The tests look quite neat, but they are also a bit hard to understand. Could you add some comments / docstrings to the test utilities to explain what they are doing and maybe link to the pytango docs to explain the magic `DeviceTestContext` is doing?

Thank you for the feedback Gary. I think I will try get this merged first and then cascade your comments through my other PRs 

Id say the core issue is that when I had the task to add a "Tango backend" - I understood that backend implys that Tango needs to do the work. Therefore I had the Tango server:
- Execute the connect method
- Poll the attributes

However, Tango is running its own event loop (I did look at the source and there is no way to pass it an eventloop, maybe theres a way to fool get_event_loop with a context manager of sorts but also afraid of the weird Tango threading) therefore I believe this is why there were issues reported with the locks.

The solution is to not poll the attributes and leave that to FastCS. The tango server is simply given the method to read the attributes - which is what gets triggered by a Tango Client request. So the server is only a thin layer. GreenMode here allows the tango server to execute the awaitable methods like the async Commands

Btw the change from MethodType for commands is because I could not get it testable (doesnt seem to raise call_count?)

Interestingly, the testing is fairly consistent with the testing suggested for FastAPI. I agree it could do with more explanation

GDYendell:
Ah OK I see, it is the wrapper functions that have changed. So, the wrapped fget now just calls `get` and not `update`. Is tango still calling that method periodically, but it is just to sync the value from the fastcs attribute into the tango device?

It might be nicer to use the attribute update_callback to call a tango device function to update the value when it changes, like the EPICS backend does [here](https://github.com/DiamondLightSource/FastCS/blob/main/src/fastcs/backends/epics/ioc.py#L171).

marcelldls:
> Ah OK I see, it is the wrapper functions that have changed. So, the wrapped fget now just calls `get` and not `update`. Is tango still calling that method periodically, but it is just to sync the value from the fastcs attribute into the tango device?
> 
> It might be nicer to use the attribute update_callback to call a tango device function to update the value when it changes, like the EPICS backend does [here](https://github.com/DiamondLightSource/FastCS/blob/main/src/fastcs/backends/epics/ioc.py#L171).

No, I believe that Tango is only reading that method when the client makes a request and then gives the result to the client. I am not aware that it records the value... According to Tom, such client polling is used at some facilities.

But this is a good point. Tom has mentioned that using Tango events is also popular in the Tango space. Hmm doesnt `update_callback` update regardless of "change"? If we can make it execute on a change only (i.e. the polled attribute was `true` and now its `false`) then we could link the Tango method `push_change_event (self, attr_name, data)` which can be used for event subscribers. Maybe this is better for a future PR as it will need more investigation and testing?

marcelldls:
@GDYendell Comments processed, tests fixed. I think this is ready

GDYendell:
> Hmm doesnt update_callback update regardless of "change"?

It does currently still call the callback if the value has not changed, but we could rethink that. softioc ignores sets with the same value and that does cause problems sometimes. A common annoyance is if you make a PV that runs a command when set to 1, then that PV needs to be manually set back to 0 afterwards for the user to be able to run the command again. We would have to be careful not to recreate the same issue at the attribute level. Let's leave it for now.

tests/conftest.py
```diff
@@ -80,11 +96,63 @@ async def counter(self):
...
+        self.mocker = mocker
+
+    @contextmanager
+    def assertPerformed(
+        self, path: list[str], action: Literal["READ", "WRITE", "EXECUTE"]
+    ):
+        """
+        This context manager can be used to confirm that a fastcs
+        controller's respective attribute or command methods are called
+        a single time witin the context block
```
GDYendell, line 111: within

marcelldls: Oops. Thanks

marcelldls:
> It does currently still call the callback if the value has not changed, but we could rethink that

This would probably also be useful for webhooks

tests/backends/tango/test_dsr.py
```diff
@@ -1,37 +1,125 @@
...
+        client_read(["BigEnum"], AttrR(Int(), allowed_values=list(range(1, 18)))._value)
+
+    def test_go(self, client_exec):
+        client_exec(["Go"])
+
+    def test_read_child1(self, client_read):
+        client_read(["SubController01", "ReadInt"], AttrR(Int())._value)
+
+    def test_read_child2(self, client_read):
+        client_read(["SubController02", "ReadInt"], AttrR(Int())._value)
```
GDYendell, line 125: On as second look, I think these tests are suffering in readability by trying to be too DRY. It isn't really clear what is being tested and I think someone trying to add new tests will either struggle to understand how to do it, or worse just copy and paste things without understanding what they are testing.

The `tango_context` fixture is good, but I think it should be used directly within each test. I think AssertableController is too much. Having a utility to recurse down the subcontrollers is useful, but I think the mocking and assertions should be done in each test.

What do you think? Maybe we could pair on this and see if we can improve it?

marcelldls: My argument is that I have been able to apply a consistent approach to other "backends", I think consistency between different backends is favorable so that we dont have to maintain bespoke things for every possible variation.  Alternatively, someone adding tests will have to understand multiple different ways. If I needed to add soft ioc tests today: I dont understand how to do it and would have to copy and paste things without understanding what I am testing because there is no structure. Also I have 600 lines to search to copy and paste from and need to have in depth knowledge of the backend to mock the right things

Happy to pair and discuss - Its not clear to me why you would divert from having a consistent way of testing

marcelldls:
@GDYendell Thanks for the additional remarks, I like it. The only thing I am not liking is the enum being "" as default, but I dont feel that is work for this PR

tests/backends/tango/test_dsr.py
```diff
@@ -90,36 +43,64 @@ def test_status(self, tango_context):
...
-    def test_read_child1(self, client_read):
-        client_read(["SubController01", "ReadInt"], AttrR(Int())._value)
-
-    def test_read_child2(self, client_read):
-        client_read(["SubController02", "ReadInt"], AttrR(Int())._value)
+    def test_read_int(self, assertable_controller, tango_context):
+        expect = 0
+        with assertable_controller.assert_read_here(["read_int"]):
+            result = tango_context.read_attribute("ReadInt").value
+        assert result == expect
```
GDYendell, line 50: nitpick: I would just put literal 0 here and remove the local variable

tests/backends/tango/test_dsr.py
```diff
@@ -90,36 +43,64 @@ def test_status(self, tango_context):
...
+
+    def test_read_write_int(self, assertable_controller, tango_context):
+        expect = 0
+        with assertable_controller.assert_read_here(["read_write_int"]):
+            result = tango_context.read_attribute("ReadWriteInt").value
+        assert result == expect
+        with assertable_controller.assert_write_here(["read_write_int"]):
+            tango_context.write_attribute("ReadWriteInt", expect)
```
GDYendell, line 58: It might be good if this did a read, a write to as different value and then another read to check it has changed

tests/backends/tango/test_dsr.py
GDYendell: I think this looks much clearer now!

tests/backends/tango/test_dsr.py
marcelldls: The reason is because:
- I want the tests to be consistent and for booleans I get  a Ruff warning "Avoid equality comparisons to `False`; use `if not result:` for false checks"
- I dont want to have `assert not result`  - its self documenting to say `assert result == expected`

tests/backends/tango/test_dsr.py
marcelldls: OK will add

marcelldls:
@GDYendell all good?

GDYendell (approved):

PR #67, opened 2024-10-29
Refactor developer API with transports and app launcher

marcelldls:
Now that im working on a device with both an epics and tango option, im finding the current public interface a bit clumsy

Particularly:
- The dot paths are awkward
- If you give Epics/Tango options, you have already enough information to describe the backend needed but you still have to import the specific backend
- If each backend will init with args like "pv_prefix" then the interfaces will never be consistent
- Backend.run calls self._run() - Therefore EpicsIOCOptions are never passed (https://github.com/DiamondLightSource/FastCS/blob/main/src/fastcs/backends/epics/backend.py#L22)
- Its not explicit what each backend should implement
- Backend specific imports are being imported regardless and there are interdependencies (tango devices get `INFO: PVXS QSRV2 is loaded, permitted, and ENABLED`) .

To think about:
-  We might consider adding `[project.optional-dependencies]` to install only what is needed in the manner done by the Polars library `pip install 'fastcs[tango]'`
- What is "asyncio backend"?
- Fix tests when happy

This is opinionated so heres a draft

Im proposing to go from:
```
from fastcs.backends.epics.backend import EpicsBackend, EpicsGUIOptions
from fastcs.connections.serial_connection import SerialConnectionSettings

backend = EpicsBackend(get_controller(port, baud), pv_prefix)
backend.create_gui(EpicsGUIOptions(output_path / "index.bob"))
backend.run()
```

To:
```
from fastcs import FastCS
from fastcs.backends import EpicsGUIOptions, EpicsIOCOptions, EpicsOptions
from fastcs.connections import SerialConnectionSettings

options = EpicsOptions(
    ioc=EpicsIOCOptions(pv_prefix=pv_prefix),
    gui=EpicsGUIOptions(output_path / "index.bob"),
)

fast_cs = FastCS(get_controller(port, baud), options)
fast_cs.create_gui()
fast_cs.run()
```
A tango instance will then look like:
```
from fastcs.backends import TangoDSROptions, TangoOptions

options = TangoOptions(
    TangoDSROptions(
        ...
    )
)

fast_cs = FastCS(get_controller(port, baud), options)
fast_cs.run()
```
Heres my example device https://github.com/DiamondLightSource/thorlabs-mff-fastcs/pull/8


marcelldls:
Considering that the Epics/Tango control system support is intended to give the external interface for the controller - is the name `frontend` not more appropriate than `backend`?

marcelldls:
I believe in the same topic there is merit in:
- A launcher that creates a consistent way to start fastcs applications. Especially so we can share a simple reusable helm chart
- Exposing a demo device

GDYendell:
I broadly agree with these changes.

It is interesting that you have repurposed the `Backend` class to just be a wrapper of the `Controller`. I think we could build on this to also merge `Mapping` and `SingleMapping` into `Backend`. I have wanted to get rid of these classes somehow for a while now because it is pretty opaque what they do and we couldn't come up with good names for them. This could be in a separate PR.

GDYendell:
> What is "asyncio backend"?

This was also difficult to name. It is basically no backend, just an interactive python shell where you can manually poke things.

marcelldls:
> > What is "asyncio backend"?
> 
> This was also difficult to name. It is basically no backend, just an interactive python shell where you can manually poke things.

I see! The soft ioc dependency confused me, I would hope to try break interdependency. Tom has supported both the idea of dropping the interactive terminal and just using a debugger but equally also the idea of an interactive terminal that works for all transport options to just poke variables. Im not sure we will be able to do an interactive terminal for Tango without solving the higher problem of running multiple transports at once because I run the server in the main thread as I dont get an option to pass an event loop.

marcelldls:
@GDYendell 
Additionally I am including a fastcs launcher (as proposed by @coretl), which can be used as a common entrypoint by using typehinting to discover the expected inputs

A minimal example:
```
from dataclasses import dataclass

from fastcs import launch
from fastcs.attributes import AttrRW
from fastcs.controller import Controller
from fastcs.datatypes import Bool


@dataclass
class TestControllerConfig:
    name: str


class TestController(Controller):
    read_write_bool: AttrRW = AttrRW(Bool())

    def startup(self, config: TestControllerConfig):
        self._config = config

    async def connect(self):
        print(f"Hello {self._config.name}")


if __name__ == "__main__":
    launch(TestController)
```

When executed:
```
$ python example.py --help
                                                                                                                                                             
 Usage: example.py [OPTIONS] COMMAND [ARGS]...                                                                                                               
                                                                                                                                                             
╭─ Options ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --install-completion          Install completion for the current shell.                                                                                   │
│ --show-completion             Show completion for the current shell, to copy it or customize the installation.                                            │
│ --help                        Show this message and exit.                                                                                                 │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Commands ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ schema   Produce json schema for a TestController                                                                                                         │
│ run      Start up a TestController                                                                                                                        │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

```
$ python example.py run --help
                                                                                                                                                                     
 Usage: example.py run [OPTIONS] CONFIG                                                                                                                              
                                                                                                                                                                     
 Start up a TestController                                                                                                                                           
                                                                                                                                                                     
╭─ Arguments ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ *    config      PATH  A yaml file matching the TestController schema [default: None] [required]                                                                  │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Options ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --help          Show this message and exit.                                                                                                                       │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

Example of the config which conforms to the expected schema (to do: handling of a k8s values.yaml)
```
# yaml-language-server: $schema=schema.yaml
controller:
  name: World
transport:
  ioc:
    terminal: true
    pv_prefix: MY-DEVICE-PREFIX
  gui:
    output_path: /home/esq51579/WIP/thorlabs-mff-fastcs/src/thorlabs_mff_fastcs/output.bob
    file_format: .bob
    title: Simple Device

```

src/fastcs/main.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def create_gui(self) -> None:
+        self._transport.create_gui()
+
+    def run(self) -> None:
+        self._backend.run()
+        self._transport.run()
+
+
+def launch(controller_class: type[Controller]):
```
GDYendell, line 57: Could this be an instance method of `FastCS`?

src/fastcs/main.py
```diff
@@ -0,0 +1,126 @@
...
+
+class FastCS:
+    def __init__(
+        self,
+        controller: type[Controller],
+        controller_options: Any,
+        transport_options: EpicsOptions | TangoOptions,
+    ):
+        self._controller = controller()
+        self._controller.startup(controller_options)
```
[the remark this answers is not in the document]
marcelldls: @GDYendell I cant find `initialise`? Mainly, I am trying to enforce a method signature where I can get the config from type hinting, I cant do this from __init__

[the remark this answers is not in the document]
GDYendell: Defined in `Controller` [here](https://github.com/DiamondLightSource/FastCS/blob/main/src/fastcs/controller.py#L58) and called by the `Backend` [here](https://github.com/DiamondLightSource/FastCS/blob/main/src/fastcs/backend.py#L26). It is used to do any initial setup that needs to be asynchronous (e.g. lots of http requests), as `__init__` can't (easily) be asynchronous.

>  I cant do this from `__init__`

Ah. Why is that?

GDYendell (commented):
Comments on first commit

tests/test_main.py
```diff
@@ -0,0 +1,127 @@
...
+    app = _launch(IsHinted)
+    result = runner.invoke(app, ["schema"])
+    assert result.exit_code == 0
+    result_dict = json.loads(result.stdout)
+
+    assert result_dict == target_dict
+
+    # # store a schema to use for debugging
+    # with open(data / "schema.json", mode="w") as f:
+    #     json.dump(result_dict, f, indent=2)
```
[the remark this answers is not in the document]
marcelldls: I did think about this. I decided to rather to keep it as unit testing and focus on the cli in isolation. We could add additional tests that check regression. A script is cleaner, I can do similar

GDYendell:
I am broadly in agreement with the changes so far, but this adds more use of "mapping" within controller, which I don't like. 

I think we should have a new class `ControllerAPI` that will replace `SingleMapping`, but as a tree (i.e. each node has its path and children: `list[ControllerAPI]`) and have some additional logic. Then we can have a concept of walking the controller API to create the transport layer and the UIs, rather than walking . Because when a `Controller` is first instantiated it is partially initialised (attributes can be added during `initialise`) I think it would make sense for the `Backend` to convert it to `ControllerAPI` that is static and then pass that to the transport layer, which should simplify the logic there. In fact, maybe `initialise` should return the `ControllerAPI`?

src/fastcs/util.py
```diff
@@ -1,5 +1,55 @@
...
+            try:
+                ret = func(*func_args)
+                if inspect.isawaitable(ret):
+                    await ret
+                if completion:
+                    completion(*completion_args)
+            except Exception:
+                logging.exception("Exception when running dispatched callback")
+
+        asyncio.run_coroutine_threadsafe(async_wrapper(), self.loop)
```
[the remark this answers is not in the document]
marcelldls: I agree with you. its not only the dependancy that bothers me but also that importing `AsyncioDispatcher` from softioc dumps `INFO: PVXS QSRV2 is loaded, permitted, and ENABLED.` (nothing to do with this class)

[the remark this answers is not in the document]
GDYendell: Added #94 to resolve this later

marcelldls:
@GDYendell Ready

src/fastcs/__init__.py
```diff
@@ -7,5 +7,7 @@
 """
 
 from ._version import __version__
+from .launch import FastCS as FastCS
+from .launch import launch as launch
```
GDYendell, line 11: Why do we need the `as`?

marcelldls: Good question. My understanding is a bit bottom up on this. Basically if I remove the `as` then you will get ruff: `src/fastcs/__init__.py:11:21: F401 .launch.launch imported but unused; consider removing, adding to '__all__', or using a redundant alias.` To solve this I looked over at https://github.com/fastapi/fastapi/blob/master/fastapi/__init__.py and this is how they do it. I dont have a PEP to give you

GDYendell: Ah it does the same as `__all__`, but neater. OK sure.

src/fastcs/transport/__init__.py
```diff
@@ -0,0 +1,8 @@
+from .epics.options import EpicsDocsOptions as EpicsDocsOptions
+from .epics.options import EpicsGUIOptions as EpicsGUIOptions
+from .epics.options import EpicsIOCOptions as EpicsIOCOptions
+from .epics.options import EpicsOptions as EpicsOptions
+from .tango.options import TangoDSROptions as TangoDSROptions
+from .tango.options import TangoOptions as TangoOptions
+
+__all__ = ["EpicsOptions", "TangoOptions"]
```
GDYendell, line 8: We don't need the `__all__` here because of the import as

src/fastcs/connections/__init__.py
```diff
@@ -1,3 +1,7 @@
 from .ip_connection import IPConnection
+from .ip_connection import IPConnectionSettings as IPConnectionSettings
+from .ip_connection import StreamConnection as StreamConnection
+from .serial_connection import SerialConnection as SerialConnection
+from .serial_connection import SerialConnectionSettings as SerialConnectionSettings
 
 __all__ = ["IPConnection"]
```
GDYendell, line 7: Could you also update this import and remove `__all__`?

src/fastcs/transport/__init__.py
marcelldls: Agreed. I also noticed I am missing "RestOptions". Messed up the rebase here

src/fastcs/__init__.py
GDYendell: Ah yes you are right. Even better. It allows it to be imported from the top-level explicitly (without unused import errors, which is the second effect of `__all__`), but without allowing to be imported as with `*`

GDYendell (approved):
Happy to squash merge because it is not clear what commits this could be broken into that would still be functional in isolation, but could you write a good summary of all the things that are being added/changed here when you do the merge?

PR #68, opened 2024-10-29
Add scheduler for scan tasks

marcelldls:
Why:
- Set intervals rather than delays (avoid drift, add precision) currently the property name `period` is a bit disingenuous
- Behavior for scan tasks that take longer than their interval. Now sets to skip - Get a useful warning message: `Execution of job "readback_position (trigger: interval[0:00:00.010000], next run at: 2024-10-29 15:13:39 GMT)" skipped: maximum number of running instances reached (1)`

Possibly useful options not used:
- Dynamically add/remove tasks (Destroy tasks on exit?)
- Pause tasks (during reconnect?)
- Custom triggers (e.g. mock trigger so that we can step in testing, Cron type triggers)
- Grace period

Uncertain:
- Handle the log messages produced

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,6 +12,7 @@ classifiers = [
 description = "Control system agnostic framework for building Device support in Python that will work for both EPICS and Tango"
 dependencies = [
     "aioserial",
+    "apscheduler",
     "numpy",
     "pydantic",
     "pvi~=0.10.0",
--- a/src/fastcs/backend.py
+++ b/src/fastcs/backend.py
@@ -1,9 +1,7 @@
 import asyncio
-from collections import defaultdict
-from collections.abc import Callable
-from concurrent.futures import Future
 from types import MethodType
 
+from apscheduler.schedulers.asyncio import AsyncIOScheduler
 from softioc.asyncio_dispatcher import AsyncioDispatcher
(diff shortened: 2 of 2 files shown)


GDYendell:
This seems good! I agree with both of your "Why"s.

I am not familiar with `apscheduler`. Is that considered the standard solution for this? Are there alternatives we should look at?

marcelldls:
Reasons for:
- Support for asyncio
- Actively maintained, 6300 stars, 300000 daily downloads (Seems to be usually used as a plugin in web applications django-apscheduler, Flask-APScheduler, etc.)
- It seems to be popular enough that there are plenty of tutorials available https://www.youtube.com/watch?v=sRnsR-T0Lxc

Possible doubts:
- Maybe too heavy a library for what we need?
- Overhead on polling performance?

An alternative could be:
- DIY
  -  Keep track of active periodic tasks and do our own logic and handling, use a call back to pop task when done
  - Use a `while <attribute>` to stop execution
  -  Move sleep into gather()? like in https://stackoverflow.com/questions/37512182/how-can-i-periodically-execute-a-function-with-asyncio
- aiojobs? 850 stars, 5000 daily downloads. Pretty minimal so possibly a middle ground

PR #69, opened 2024-11-04
First attempt rest api

marcelldls:
An additional "backend" for FastCS which provides a RestAPI.
The API is fully discoverable and demoable from the autogenerated "Swagger gui" at "{domain}/docs" and the schema can be retrived from "{domain}/openapi.json", all thanks to FastAPI.

The following considersations have been taken to conform with restful design:
- Attributes can be read using a GET request to "{domain}/{attribute-name}" (note if sub-controller then {domain}/{subcontroller name}/{attribute-name}" etc.)
  - A successful request receives response of status code 200 and a payload of json: {"value": value}
- Attributes can be written using a PUT request to "{domain}/{attribute-name}" etc.
  - A PUT request is used rather than a GET of "{domain}/{attribute-name}/{value}" to not be triggered accidentally (not "safe")
  - A PUT request is used over a POST as the action is idempotent
  - The request would have a payload of json: {"value": value}
  - A 204 response to confirm the request as there is no information to return
  - FastAPI automatically deals with wrong types an will issue a 422 error with message
- Commands can be triggered using an empty PUT to "{domain}/{command-name}"
  - A 204 response to confirm the request as there is no information to return

I follow a similar pattern to the Tango "backend" to create the server and do some testing, so any feedback from #66 should feed into this


src/fastcs/backends/rest/rest.py
```diff
@@ -0,0 +1,147 @@
...
+            host=options.host,
+            port=options.port,
+            log_level=options.log_level,
+        )
+
+
+def _put_request_body(attribute: AttrW[T]):
+    return create_model(
+        f"Put{str(attribute.datatype.dtype)}Value",
+        **{"value": (attribute.datatype.dtype, ...)},  # type: ignore
```
GDYendell, line 49: This could do with some explanation. And the two similar functions below.

Why is the key `value`? What does `(attribute.datatype.dtype, ...)` do? Maybe just a case of referencing FastAPI docs?

src/fastcs/backends/rest/rest.py
```diff
@@ -0,0 +1,147 @@
...
+    attribute: AttrR[T],
+) -> Callable[[], Coroutine[Any, Any, Any]]:
+    async def attr_get() -> Any:  # Must be any as response_model is set
+        value = attribute.get()  # type: ignore
+        return {"value": value}
+
+    return attr_get
+
+
+def _add_dev_attributes(app: FastAPI, mapping: Mapping) -> None:
```
GDYendell, line 82:
```suggestion
def _add_attribute_api_routes(app: FastAPI, mapping: Mapping) -> None:
```

src/fastcs/backends/rest/rest.py
```diff
@@ -0,0 +1,147 @@
...
+def _wrap_command(
+    method: Callable, controller: BaseController
+) -> Callable[..., Awaitable[None]]:
+    async def command() -> None:
+        await getattr(controller, method.__name__)()
+
+    return command
+
+
+def _add_dev_commands(app: FastAPI, mapping: Mapping) -> None:
```
GDYendell, line 132:
```suggestion
def _add_command_api_routes(app: FastAPI, mapping: Mapping) -> None:
```

src/fastcs/backends/rest/rest.py
```diff
@@ -0,0 +1,147 @@
...
+            attr_name = attr_name.title().replace("_", "")
+            d_attr_name = f"{'/'.join(path)}/{attr_name}" if path else attr_name
+
+            match attribute:
+                # https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods
+                case AttrRW():
+                    app.add_api_route(
+                        f"/{d_attr_name}",
+                        _wrap_attr_get(attribute),
+                        methods=["GET"],  # Idemponent and safe data retrieval,
```
GDYendell, line 96: Typo Idemponent

src/fastcs/backends/rest/rest.py
```diff
@@ -0,0 +1,147 @@
...
+    return attr_get
+
+
+def _add_dev_attributes(app: FastAPI, mapping: Mapping) -> None:
+    for single_mapping in mapping.get_controller_mappings():
+        path = single_mapping.controller.path
+
+        for attr_name, attribute in single_mapping.attributes.items():
+            attr_name = attr_name.title().replace("_", "")
+            d_attr_name = f"{'/'.join(path)}/{attr_name}" if path else attr_name
```
GDYendell, line 88: Maybe `route` or `path` would be a better name here?

GDYendell (commented):
Looks good. A few comments.

I assume the test utilities have just been copied from the tango PR for now and most of it can be shared between backends once merged?

marcelldls:
> Looks good. A few comments.
> 
> I assume the test utilities have just been copied from the tango PR for now and most of it can be shared between backends once merged?

Thank you for the review Gary. I found a few tricks when I did the web api testing and fed this back into Tango. Since we will add p4p I did not look at softioc...

src/fastcs/backends/rest/rest.py
marcelldls: > Why is the key `value`? 

My intention was a schema defined for the request body for each datatype to have the least possible schemas. An alternative could be a schema with the attribute as a key, but thats potentially a lot of schemas that are the same and less navigable in `/docs`.

I choose "value" because im not aware of standards for this but it seems fairly common (as per https://learn.microsoft.com/en-us/azure/azure-app-configuration/rest-api-key-value?pivots=v1). I dont want to use "data" because that seems to be often used as an object ( as per https://developer.hashicorp.com/vault/api-docs).

> What does `(attribute.datatype.dtype, ...)` do?

This creates the schema without defaults. I will add a comment

src/fastcs/backends/rest/rest.py
marcelldls: Agree

src/fastcs/backends/rest/rest.py
marcelldls: Good catch

src/fastcs/backends/rest/rest.py
marcelldls: I avoided "path" because its already used in this context. I avoided "route" because its not the full route (missing the root `/`) but if it makes it clearer I will add it

src/fastcs/backends/rest/rest.py
GDYendell: Also, I think you can remove the `# type: ignore` if you do 

```
return create_model(
    f"Put{str(attribute.datatype.dtype)}Value",
    value=(attribute.datatype.dtype, ...),
)
```

?

marcelldls: > Also, I think you can remove the `# type: ignore` if you do
> 
> ```
> return create_model(
>     f"Put{str(attribute.datatype.dtype)}Value",
>     value=(attribute.datatype.dtype, ...),
> )
> ```
> 
> ?

This is true - Thank you

src/fastcs/backends/rest/rest.py
marcelldls: I have also realised that kebab case seems to be best practice for a restapi - I can do this for the attribute/command level. But controller/subcontroller name I will leave for the developer?

src/fastcs/backends/rest/rest.py
marcelldls: I have noticed that `f"Put{str(attribute.datatype.dtype)}Value"` renders out as `Put<class 'bool'>Value`. I think if I were to be consistent with my intention of a schema per datatype I should aim to create `PutBoolValue`

marcelldls:
@GDYendell Updated the testing inline with feedback from #66

src/fastcs/backends/rest/rest.py
```diff
@@ -0,0 +1,158 @@
...
+        for name, method in single_mapping.command_methods.items():
+            cmd_name = name.replace("_", "-")
+            route = f"/{'/'.join(path)}/{cmd_name}" if path else cmd_name
+            app.add_api_route(
+                f"/{route}",
+                _wrap_command(
+                    method.fn,
+                    single_mapping.controller,
+                ),
+                methods=["PUT"],
```
GDYendell, line 156: Should commands use POST?

GDYendell: Ah

> A PUT request is used over a POST as the action is idempotent

Could you expand on this?

marcelldls: Good to discuss

https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/POST
```
The difference between [PUT] and POST is that PUT is idempotent, 
calling it once is no different from calling it several times successively (there are no side effects). 
Successive identical POST requests may have additional effects, 
such as creating the same order several times.
```

Here a more concrete example:

https://www.theserverside.com/blog/Coffee-Talk-Java-News-Stories-and-Opinions/PUT-vs-POST-Whats-the-difference
```
For example, I perform a PUT operation to set my bank account balance to $100. 
Even if I perform that operation 100 times, the end result is a balance of $100.

In contrast, if I perform 100 POST operations that each increases my account balance by 10%, 
the result of each interaction creates a different result. That is not an idempotent operation.

A PUT method allows only idempotent operations. 
POST methods are not subject to idempotency requirements."
```
By this definition I consider a command is a PUT

A POST could be "safe" in a sense that a command may not be idempotent. A big distinction seems to be that a POST must return a body - which is not something are doing in EPICS...

marcelldls: Good timing! Fair enough. Of course I will be biased but I like having `/subcontroller/{method or attribute}` as the route so that we can see what is available from the docs and having very simple schemas and logic. Do you have other suggestions?

GDYendell: Is there a precedent in FastAPI that we can copy here?

marcelldls: FastAPI does not constrain you to using a RestAPI, so we should just try be restful which is why I refer to the mozilla docs. An interesting reference is possibly "https://tango-controls.readthedocs.io/en/latest/development/advanced/rest-api.html" they do use an empty put for commands, but they support 1 arg

marcelldls: Note the API we use is fully described at `/docs` by swagger

GDYendell: OK. With odin I think we are just going with PUT as well, so that seems reasonable.

PR #71, opened 2024-11-06
Attempt graphql backend

marcelldls:
An additional "backend" for FastCS which provides a GraphQL api.
The API is fully discoverable and demoable from the autogenerated "GraphiQL gui" at "{domain}/graphql".
I choose to use the FastAPI "Strawberry" integration as it is the recommended library by FastAPI.

The following considersations have been taken to conform with best practices:
- Attributes to be read using a query (also demonstates how subcontrollers are nested)

Example:
```
Query request:
query{
    DesiredPosition
    subcontroller {
        DesiredPosition
    }
}

Query response:
{
  "data": {
    "subcontroller": {
      "DesiredPosition": 1.0,
    }
    "DesiredPosition": 1.0,
  }
}
```

- Attributes to be written using a mutation 

Example:
```
Mutation request:
mutation{
  DesiredPosition(value: 2.0)
}

Mutation response:
{
  "data": {
    "DesiredPosition": 2.0
  }
}
```
- Commands can cause server side effects so they are triggered using a mutation as per convention
  - A boolean response is returned to confirm the command was executed successfully

Example:
```
Mutation request:
mutation{
  BlinkLED
}

Mutation response:
{
  "data": {
    "BlinkLED": true
  }
}
```

Further:
- I follow a similar pattern to the Tango "backend" to create the server and testing, so any feedback from the exisiting PRs should feed into this
- I have added nested subcontrollers to test arbitrary subcontroller nesting as this is a harder problem for the graphql request structure - the tests of other backends should also apply this and need to be updated accordingly


GDYendell:
I have a had a quick look, but I am going to ignore this until we iterate on and merge some of the other recent PRs first, as you suggest.

marcelldls:
> I have a had a quick look, but I am going to ignore this until we iterate on and merge some of the other recent PRs first, as you suggest.

Lets do that. When we are done iterating, we can enjoy some of the graph recursion xD

marcelldls:
Apparently camelCase is the convention for field names https://www.apollographql.com/docs/graphos/schema-design/guides/naming-conventions#casing so i have adjusted accordingly

src/fastcs/backends/graphQL/graphQL.py
```diff
@@ -0,0 +1,225 @@
...
+                    new_type,
+                )
+                self.fields_dict[strawberry_type].append(child_field)
+
+        if self.fields_dict[strawberry_type]:
+            return create_type(
+                f"{self.name}{strawberry_type}", self.fields_dict[strawberry_type]
+            )
+        else:
+            return None
```
GDYendell, line 150: I am quite confused by this method. I don't understand when it would return `None`. Isn't `strawberry_type` always just `"query"` or `"mutation"`?

src/fastcs/backends/graphQL/graphQL.py
```diff
@@ -0,0 +1,225 @@
...
+        for child in self.children:
+            if child.name == name:
+                return True
+        return False
+
+    def get_child(self, name: str) -> "FieldsTree":
+        for child in self.children:
+            if child.name == name:
+                return child
+        raise NodeNotFoundError
```
GDYendell, line 134: These two methods would be neater if children was a dict not a list.

GDYendell (commented):
First pass comments

src/fastcs/backends/graphQL/graphQL.py
```diff
@@ -0,0 +1,230 @@
...
+        if path:
+            node = fields_tree.insert(path)
+        else:
+            node = fields_tree
+
+        if node is not None:
+            for attr_name, attribute in single_mapping.attributes.items():
+                # camelCase is convension for field names
+                attr_name = attr_name.title().replace("_", "")
+                attr_name = attr_name[0].lower() + attr_name[1:]
```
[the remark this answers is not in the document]
marcelldls: Amazing! Nice catch. Didn't think it would be so clever

src/fastcs/backends/graphQL/graphQL.py
```diff
@@ -0,0 +1,225 @@
...
+        name = path.pop(0)
+        if self.is_child(name):
+            child = self.get_child(name)
+        else:
+            child = FieldsTree(name)
+            self.children.append(child)
+
+        # Recurse if needed
+        if path:
+            return child.insert(path)  # type: ignore
```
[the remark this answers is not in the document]
marcelldls: Good one. Thanks

src/fastcs/backends/graphQL/graphQL.py
marcelldls: Agree

marcelldls: I think I should actually just have `get_child` return `FieldsTree | None` and get rid of `is_child`. I think at the time I was using `is_child` in more places but now it only gets used to validate before `get_child` is called

src/fastcs/backends/graphQL/graphQL.py
```diff
@@ -0,0 +1,225 @@
...
+    async def _dynamic_f() -> bool:
+        await getattr(controller, method.__name__)()
+        return True
+
+    _dynamic_f.__name__ = method_name
+
+    return _dynamic_f
+
+
+def _add_dev_commands(
```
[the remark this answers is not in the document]
marcelldls: thanks

marcelldls:
I will rebase this on monday

marcelldls:
@GDYendell Rebased. Also did the rename of create_type as discussed this morning

GDYendell (approved):
Tenuous self approval...

image

PR #72, opened 2024-11-11
Panda conversion improvements

evalott100:
# Changes

### Mapping
Controllers can now define
``` python
    @property
    def additional_attributes(self) -> dict[str, Attribute]:
        """FastCS will look for attributes on the controller, but additional attributes
        are provided by this method."""
        return {}
```

which is another way to provide attributes without `setattr` on the controller. The keys are checked to be distinct from the attributes found from the attributes of the controller.

The user can also pass in `search_device_for_attributes=False` on the controller which will prevent the mapping automatically searching for `Attributes`, it will only use `additional_attributes`.

### Description
Added a field called `description` to `Attribute`, which is used as `DESC` on the `EPICS` record.

### Initial value
Added an option for `initial_value` on `SignalR`/`SignalRW`. If provided this will be used instead of the default of the `_datatype.dtype` for the initial value of the attribute.

### Changing the datatype
We still want `DataType` to be frozen, but we also might want to change the value of a field in it - e.g `units`, corresponding to `EGU` in the epics backend.

To do this I've made a method called `update_datatype` which will change the `_datatype` of the attribute. This alone wouldn't update the corresponding record, so the backend runs `set_update_databack_callback`. `update_datatype` will then run this so changing the attribute datatype will update the backend.

### Errors
Currently, errors in a scan future will be swallowed. Now an error will be printed.

### Killing Scan Tasks
We were getting problems for scan tasks not being torn down after the end of tests. Now the backend will stop the tasks on `__del__`, or running the new `stop_scan_tasks` method.

### Methods
Rather than just using `DataType.dtype()` we now have a `DataType.initial_value` property.
We also have a `DataType.cast` which is used to cast inputted values and validated against metadata fields in the datatype.

### Added metadata to the datatype
Added a bunch of optional fields to `String, Int, Float` which are used for relevant fields e.g `HOPR` in the epics backend. `Int` and `Float` `Attributes` are validated against their `max` and `min` on put.

#### Waveform
A waveform to correspond to a single pv without any pva structure beyond regular pvi.

#### Table
A waveform which is indexable using pvi PVs generated from names of the columns of the numpy structured datatype. These will be added in a new PVXS backend.

### Options
Naming conventions are now passed into the ioc through `EpicsIOCOptions`, which can be passed into the backend on init.

### Poll Period
`update_period` is now optional in fields, designating that the `Updater` shouldn't be ran periodically. In panda we get all the changes at once in the top level `update` scan method. We want to then update all the relevant attributes from there.

### DIFF
--- a/src/fastcs/attributes.py
+++ b/src/fastcs/attributes.py
@@ -27,7 +27,7 @@ class Updater(Protocol):
     """Protocol for updating the cached readback value of an ``Attribute``."""
 
     # If update period is None then the attribute will not be updated as a task.
-    update_period: float | None
+    update_period: float | None = None
 
     async def update(self, controller: Any, attr: AttrR) -> None:
         pass
@@ -95,6 +95,7 @@ class AttrR(Attribute[T]):
         access_mode=AttrMode.READ,
         group: str | None = None,
         handler: Updater | None = None,
+        initial_value: T | None = None,
         allowed_values: list[T] | None = None,
         description: str | None = None,
     ) -> None:
@@ -106,7 +107,7 @@ class AttrR(Attribute[T]):
             allowed_values=allowed_values,  # type: ignore
             description=description,
         )
-        self._value: T = datatype.dtype()
+        self._value: T = datatype.dtype() if initial_value is None else initial_value
         self._update_callback: AttrCallback[T] | None = None
         self._updater = handler
 
@@ -177,7 +178,7 @@ class AttrW(Attribute[T]):
         return self._sender
 
 
-class AttrRW(AttrW[T], AttrR[T]):
+class AttrRW(AttrR[T], AttrW[T]):
     """A read-write ``Attribute``."""
 
     def __init__(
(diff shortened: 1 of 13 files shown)


evalott100:
Scan tasks are not ending correctly in the tests on main:

========================================================== 30 passed in 2.93s ===========================================================
INFO: PVXS QSRV2 is loaded, permitted, and ENABLED.
Task was destroyed but it is pending!
task: <Task pending name='Task-11' coro=<_create_periodic_scan_task.<locals>.scan_task() done, defined at /home/eva/Code/FastCS/src/fastcs/backend.py:156> wait_for=<Future pending cb=[Task.task_wakeup()]> cb=[_chain_future.<locals>._call_set_state() at /usr/lib64/python3.11/asyncio/futures.py:394]>
Task was destroyed but it is pending!
task: <Task pending name='Task-12' coro=<_create_periodic_scan_task.<locals>.scan_task() done, defined at /home/eva/Code/FastCS/src/fastcs/backend.py:156> wait_for=<Future pending cb=[Task.task_wakeup()]> cb=[_chain_future.<locals>._call_set_state() at /usr/lib64/python3.11/asyncio/futures.py:394]>

GDYendell:
This is quite a lot. It would be good to split this into smaller PRs. I think some of it can be merged now and then we can continue developing and discussing the rest. It will also help resolve conflicts with the various PRs @marcelldls is working on in smaller chunks.

evalott100:
> This is quite a lot. It would be good to split this into smaller PRs. I think some of it can be merged now and then we can continue developing and discussing the rest.

I agree, I had a philosophy of putting all the changes together since I've amended them so much. Let me know which features you want now and I can cherry pick them into their own issues/PRs

evalott100:
Closing as these have all been separated out into separate PRs.

PR #77, opened 2024-11-18
76 pass backend options on init and provide pv name formatting options

### LINKED ISSUE #76: Pass backend options on init instead of run

evalott100:
Closes #76

### DIFF
--- a/src/fastcs/attributes.py
+++ b/src/fastcs/attributes.py
@@ -26,7 +26,8 @@ class Sender(Protocol):
 class Updater(Protocol):
     """Protocol for updating the cached readback value of an ``Attribute``."""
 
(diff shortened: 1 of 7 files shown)


GDYendell:
There is a lot more in this that the name suggests. Most of the changes are related to adding the naming convention config, could that be taken out? Is it tightly coupled somehow?

evalott100:
I accidentally included

`made update_period optional on Updater` [722cc37](https://github.com/DiamondLightSource/FastCS/pull/77/commits/722cc37b775a43805b03e60e04401f11aa06aa88)

Here. I dropped it from the branch but it's not been removed from this PR for whatever reason.

PR #82, opened 2024-11-20
81 add features to attributes and datatypes

### LINKED ISSUE #81: Add features to `Attributes` and `DataTypes`
The following should be added/modified in `Attribute`s: and `DataType`s:

### Attributes
* The option for an `initial_value` to `AttrR` and `AttrRW`.
* Optional `description` field.

### DataType
* Minimum and maximum values, and minimum and maximum alarm bounds, and `units` on `Float` and `Int`.
* Initial value (unprovided) on an attribute will now come from the property `initial_value` on the `DataType`.

### Changing the datatype
We still want `DataType` to be frozen, but we also might want to change the value of a field in it - e.g `units`, corresponding to `EGU` in the epics backend.

We use a new method `Attribute::update_datatype` which will change the `_datatype` of the attribute. This alone wouldn't update the corresponding record, so the backend runs `add_update_databack_callback` when it parses the attributes. `update_datatype` will then run this so changing the attribute datatype will update the backend.

### Misc
`update_period` on `Updater` is not optional. `None` won't set up a task to poll it.

evalott100:
Closes #81

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ dependencies = [
     "pydantic",
     "pvi~=0.10.0",
     "pytango",
-    "softioc",
+    "softioc>=4.5.0",
 ]
 dynamic = ["version"]
 license.file = "LICENSE"
--- a/src/fastcs/attributes.py
+++ b/src/fastcs/attributes.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from collections.abc import Callable
 from enum import Enum
 from typing import Any, Generic, Protocol, runtime_checkable
 
@@ -65,6 +66,10 @@ class Attribute(Generic[T]):
         self._allowed_values: list[T] | None = allowed_values
         self.description = description
 
+        # A callback to use when setting the datatype to a different value, for example
+        # changing the units on an int. This should be implemented in the backend.
+        self._update_datatype_callbacks: list[Callable[[DataType[T]], None]] = []
+
     @property
     def datatype(self) -> DataType[T]:
         return self._datatype
@@ -84,6 +89,20 @@ class Attribute(Generic[T]):
     @property
     def allowed_values(self) -> list[T] | None:
         return self._allowed_values
+
+    def add_update_datatype_callback(
+        self, callback: Callable[[DataType[T]], None]
+    ) -> None:
+        self._update_datatype_callbacks.append(callback)
+
+    def update_datatype(self, datatype: DataType[T]) -> None:
+        if not isinstance(self._datatype, type(datatype)):
+            raise ValueError(
+                f"Attribute datatype must be of type {type(self._datatype)}"
+            )
+        self._datatype = datatype
+        for callback in self._update_datatype_callbacks:
+            callback(datatype)
 
 
 class AttrR(Attribute[T]):
--- a/src/fastcs/attributes.py
+++ b/src/fastcs/attributes.py
@@ -27,7 +27,7 @@ class Updater(Protocol):
     """Protocol for updating the cached readback value of an ``Attribute``."""
 
     # If update period is None then the attribute will not be updated as a task.
-    update_period: float | None
+    update_period: float | None = None
 
     async def update(self, controller: Any, attr: AttrR) -> None:
         pass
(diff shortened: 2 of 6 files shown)


pyproject.toml
```diff
@@ -16,7 +16,8 @@ dependencies = [
     "pydantic",
     "pvi~=0.10.0",
     "pytango",
-    "softioc",
+    # This is needed for softioc.device.set_field
```
GDYendell, line 19: Move this note to the commit message.

src/fastcs/backends/epics/ioc.py
```diff
@@ -27,6 +27,25 @@ class EpicsIOCOptions:
...
+    "znam": "ZNAM",
+    "onam": "ONAM",
+}
+
+
+def datatype_to_epics_fields(datatype: DataType) -> dict[str, Any]:
+    return {
+        DATATYPE_NAME_TO_RECORD_FIELD[field]: value
+        for field, value in asdict(datatype).items()
+    }
```
GDYendell, line 46: This will raise a confusing error if a field is missing. Could we catch the exception and provide a useful message?

src/fastcs/datatypes.py
```diff
@@ -60,3 +73,15 @@ class String(DataType[str]):
...
+def validate_value(datatype: DataType[T], value: T) -> T:
+    """Validate a value against a datatype."""
+
+    if isinstance(datatype, (Int | Float)):
+        assert isinstance(value, (int | float)), f"Value {value} is not a number"
+        if datatype.min is not None and value < datatype.min:
+            raise ValueError(f"Value {value} is less than minimum {datatype.min}")
+        if datatype.max is not None and value > datatype.max:
+            raise ValueError(f"Value {value} is greater than maximum {datatype.max}")
+    return value
```
GDYendell, line 87: I think this should be a method of `Datatype` that with default implementation `return value`.

GDYendell (commented):
A couple of comments, but looks good.

src/fastcs/backends/epics/ioc.py
```diff
@@ -224,33 +266,62 @@ async def async_write_display(value: T):
...
+                pv, always_update=True, on_update=on_update, **attribute_fields
+            )
         case _:
             raise FastCSException(
                 f"Unsupported type {type(attribute.datatype)}: {attribute.datatype}"
             )
 
+    def datatype_updater(datatype: DataType):
+        for name, value in datatype_to_epics_fields(datatype).items():
+            record.set_field(name, value)
```
GDYendell, line 320: I think this is important to get test coverage on.

evalott100: I added a test in the squash.

evalott100: But I'll extend it for all relevant `AttR` types.

src/fastcs/backends/epics/ioc.py
evalott100: I'll make it so that it doesn't add anything if the `field` isn't defined with an epics name in the dictionary. It may only be possible to implement some fields in some backends.

src/fastcs/datatypes.py
```diff
@@ -20,9 +21,40 @@ class DataType(Generic[T]):
     def dtype(self) -> type[T]:  # Using property due to lack of Generic ClassVars
         pass
 
+    @abstractmethod
+    def cast(self, value: T) -> Any:
```
GDYendell, line 25: I don't think this should be called `cast` if `value` has to already be the correct type. In which case I am not sure the docstring is accurate.

I think all this is intended to do is check against its fields, e.g. `min`, `max`? Maybe `validate` would be more appropriate?

src/fastcs/attributes.py
```diff
@@ -170,7 +199,7 @@ def sender(self) -> Sender | None:
         return self._sender
 
 
-class AttrRW(AttrW[T], AttrR[T]):
+class AttrRW(AttrR[T], AttrW[T]):
```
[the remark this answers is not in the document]
evalott100: Passes the `initial_value` to `AttR` instead of `AttrW` on `super().__init__`

src/fastcs/datatypes.py
evalott100: For now I'll change it's name to `validate` (see below).

GDYendell (approved):

GDYendell (approved):

PR #84, opened 2024-11-20
76 pass backend options on init instead of run and provide pv name formatting options

### LINKED ISSUE #76: Pass backend options on init instead of run

evalott100:
Closes #76

Reopening because #77 was struggling with fetching a dropped commit.

### DIFF
--- a/src/fastcs/backends/epics/backend.py
+++ b/src/fastcs/backends/epics/backend.py
@@ -7,11 +7,17 @@ from .ioc import EpicsIOC, EpicsIOCOptions
 
 
 class EpicsBackend(Backend):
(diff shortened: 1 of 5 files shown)


GDYendell:
I would like to have the pv naming in a separate PR. I am not sure it is quite right yet and maybe we should rework the way we handle combinations of controller paths and attributes entirely. I don't like that you have to call `_convert_attribute_name_to_pv_name` with an `is_attribute` flag.

Also, I don't think we need  `CAPITALIZED_CONTROLLER_PASCAL_ATTRIBUTE` and it seems to be complicating the logic here a lot. I think what need need is for capitalised acronyms to be left alone regardless of the mode, e.g. FP in controller path or PVI as the leaf. PascalCase doesn't handle acronyms explicitly, but I think we should add some handling around that.

PR #88, opened 2024-11-20
pass ioc options to `EpicsIOC.__init__` instead of `run`

### LINKED ISSUE #76: Pass backend options on init instead of run

evalott100:
Partially closes #76

### DIFF
--- a/src/fastcs/backends/epics/backend.py
+++ b/src/fastcs/backends/epics/backend.py
@@ -7,11 +7,16 @@ from .ioc import EpicsIOC, EpicsIOCOptions
 
 
 class EpicsBackend(Backend):
-    def __init__(self, controller: Controller, pv_prefix: str = "MY-DEVICE-PREFIX"):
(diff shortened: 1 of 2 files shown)


src/fastcs/backends/epics/backend.py
```diff
@@ -7,17 +7,23 @@
...
+        controller: Controller,
+        pv_prefix: str = "MY-DEVICE-PREFIX",
+        options: EpicsIOCOptions | None = None,
+    ):
         super().__init__(controller)
 
         self._pv_prefix = pv_prefix
-        self._ioc = EpicsIOC(pv_prefix, self._mapping)
+        options = options or EpicsIOCOptions()
+        self._ioc = EpicsIOC(pv_prefix, self._mapping, options=options)
```
GDYendell, line 20: `EpicsIOC` already handles this so we can just pass through `EpicsIOCOptions | None` here.

PR #96, opened 2024-12-03
Add optional version command to launcher

marcelldls:
`version` is a useful cli command - this option adds the command if the developer passes a version that should be output when executing `${my_fastcs_device} version`

### DIFF
--- a/src/fastcs/launch.py
+++ b/src/fastcs/launch.py
@@ -77,7 +77,9 @@ def launch(
 
     Args:
         controller_class (type[Controller]): The FastCS Controller to instantiate.
-        It must have a type-hinted __init__ method and no more than 2 arguments.
+            It must have a type-hinted __init__ method and no more than 2 arguments.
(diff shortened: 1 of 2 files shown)


GDYendell:
Would it be possible to print the FastCS version as well as the controller version?

marcelldls:
Thats a good idea

marcelldls:
The default is that it will print FastCS version. If a version of the implemented controller is passed, it will also be shown
```
(.venv) [esq51579@pc0146 FastCS]$ thorlabs-mff-fastcs version
INFO: PVXS QSRV2 is loaded, permitted, and ENABLED.
ThorlabsMFF: my_version
FastCS: 0.7.1.dev19+gdc6f1a0
```

GDYendell (approved):
Looks good.

Note if we move towards having a single FastCS application entry point then it will need to import the controller and could potentially get the controller version itself.

PR #102, opened 2024-12-09
Add enum and waveform values

evalott100:
I won't be adding these to GraphQL in the same PR.

I'd like a quick review before I add tests for the remaining uncovered areas since it'll probably change around. @garryod  @marcelldls


GDYendell:
Wrong Gar(r)y tagged @evalott100 ?

jsouter:
Very mad that you replied Garry!!
Looks good at a first look, I'll try and run this against fastcs-odin or eiger. I can't remember if there's an issue for this in PVI or somewhere else, but I think for enums the ComboBox may only be set up to send the string value instead of the index to the handler, for odin we would have to find the associated index for that value in the Sender's put and send that via the http connection, which may be okay, depends on if we generally want to be able to have the dropdowns with different labels to the actual value that gets sent.

evalott100:
> Wrong Gar(r)y tagged in the PR description ?

Apologies this is what I get for tying `@G`

GDYendell:
> depends on if we generally want to be able to have the dropdowns with different labels to the actual value that gets sent

Yes we do. The user wants to see the text label and doesn't care if the device gets sent the label or the index.

evalott100:
> Yes we do. The user wants to see the text label and doesn't care if the device gets sent the label or the index.

This confuses me, the phoebus screen will be updating the IOC values (which would have an mbb record which takes indices or string), which will update the record through [`ParamTreeHandler::update`](https://github.com/DiamondLightSource/fastcs-odin/blob/7530b8d524268e1993276761f7078a6288af46c3/src/fastcs_odin/odin_adapter_controller.py#L43).

What would need to change is [where we make the attribute](https://github.com/DiamondLightSource/fastcs-odin/blob/7530b8d524268e1993276761f7078a6288af46c3/src/fastcs_odin/odin_adapter_controller.py#L195-L212).

`allowed_values` is now an field on the `DataType` so if

1. You want a simple string with allowed values (what you have now), then [`types`](https://github.com/DiamondLightSource/fastcs-odin/blob/7530b8d524268e1993276761f7078a6288af46c3/src/fastcs_odin/odin_adapter_controller.py#L15) should map the datatype classes instead of instances and you can initialise them in `_create_attributes` with the `allowed_values`.

This would have the exact same behaviour across the stack so long as we change

https://github.com/DiamondLightSource/FastCS/blob/cdaf83a14b564c6bf7be2f0e108b40f7e25abeec/src/fastcs/transport/epics/gui.py#L67-L72

so that strings with allowed values also make `ComboBox` (I would be in favour of this).

2. You want to use an `Enum`, then you create an enum at Runtime based on the labels given then you make quick `cast_to_http`/`cast_from_http` which get the name/value of the attribute's enum datatype and use them [here](https://github.com/DiamondLightSource/fastcs-odin/blob/7530b8d524268e1993276761f7078a6288af46c3/src/fastcs_odin/odin_adapter_controller.py#L36) and [here](https://github.com/DiamondLightSource/fastcs-odin/blob/7530b8d524268e1993276761f7078a6288af46c3/src/fastcs_odin/odin_adapter_controller.py#L57) respectively.

jsouter:
> strings with allowed values also make `ComboBox` (I would be in favour of this). 

+1
> you create an enum at Runtime based on the labels given then you make quick `cast_to_http`/`cast_from_http` which get the name/value of the attribute's enum datatype and use them [here](https://github.com/DiamondLightSource/fastcs-odin/blob/7530b8d524268e1993276761f7078a6288af46c3/src/fastcs_odin/odin_adapter_controller.py#L36) and [here](https://github.com/DiamondLightSource/fastcs-odin/blob/7530b8d524268e1993276761f7078a6288af46c3/src/fastcs_odin/odin_adapter_controller.py#L57) respectively. 

I did try this at some point in a stale branch and it worked fine so we could keep all this logic inside fastcs-odin, though would be nice if it was resolved higher up. I'm not sure if Odin is unusual in how it handles enums though I suspect there are other potential use cases that would expect ints. (odin-control's ParameterTree doesn't actually properly implemented string labels for enums in its own allowed_values/metadata currently so I was manually adding in these labels for testing, I want to rework that at some point anyway, maybe this should be more of an Odin change)

src/fastcs/transport/epics/ioc.py
```diff
@@ -243,23 +239,13 @@ def datatype_updater(datatype: DataType):
...
+    async def on_update(value):
+        await attribute.process_without_display_update(cast_method(value))
 
-        async def on_update(value):
-            await attribute.process_without_display_update(value)
-
-        async def async_write_display(value: T):
-            record.set(value, process=False)
+    async def async_write_display(value: T):
+        record.set(cast_method(value), process=False)
```
GDYendell, line 248: This should be casting _to_ an EPICS type, but this is using the same method as `on_update`.

src/fastcs/transport/epics/util.py
```diff
@@ -25,75 +28,71 @@
...
-        return attribute.allowed_values[index]
-    except IndexError:
-        raise IndexError(
-            f"Invalid index {index} into allowed values: {attribute.allowed_values}"
-        ) from None
+            def cast_from_epics_type(value) -> T:
+                return datatype.validate(value)
+        case _:
+            raise ValueError(f"Unsupported datatype {datatype}")
+    return cast_from_epics_type
```
GDYendell, line 98: I am quite confused what these methods are doing. They aren't equivalent to the previous
`enum_value_to_index` and `enum_index_to_value` functions. Why do they convert between
`T` and `object` and not `T` and `int`? Also why do they both just call `validate`,
shouldn't one of them be doing the opposite?

I think it was much clearer before when these conditionally defined functions were
inline in `ioc.py` and the cast methods were simple.

GDYendell (commented):
Initial comments

src/fastcs/transport/epics/util.py
evalott100: > Why do they convert between `T` and `object` and not `T` and `int`

Purely for typing, these methods needs to return a callable which performs the cast for a specific `DataType`. There's no way to say, if `T` is `StrEnum` return a callable to `str`, if `T` is `IntEnum` return a callable to `int`.

> Also why do they both just call `validate`, shouldn't one of them be doing the opposite

What do you mean by doing the opposite? I wanted to validate stuff both ways.

* We need to validate the value received from EPICS since it might not fit the FastCS exclusive constraints - e.g `allowed_values` on a waveform. It'll also just makes debugging a lot easier if theres a situation where a backend and `DataTypes` disagree on definitions.
* We need to validate the value sent to EPICS since it won't error if it can't accept a value (because of field constraints).

> I think it was much clearer before when these conditionally defined functions were
inline in ioc.py and the cast methods were simple.

I had a plan to structure it this way so that this kind of logic can be more easily found for each transport. I agree though, I'll put it back where it was.

GDYendell:
@evalott100 we have merged #98 - could you resolve the conflicts? I would like to get this in next.

src/fastcs/transport/epics/util.py
```diff
@@ -25,75 +31,104 @@
...
-    Raises:
-        ValueError: If `attribute` has no allowed values or `value` is not a valid
-            option
 
-    """
-    if attribute.allowed_values is None:
-        raise ValueError(
-            "Cannot convert value to index for Attribute without allowed values"
-        )
+def get_callable_to_epics_type(datatype: DataType[T]) -> Callable[[T], object]:
```
GDYendell, line 97: I think these methods would be simpler if we pass the datatype along with the value and always do the match on get/put.

GDYendell (approved):

PR #105, opened 2024-12-10
Add SimpleHandler to make minimal Attributes functional

### LINKED ISSUE #101: Attribute._value is not updated with the PV
Internal `.get()` on values just return the initial value

## Steps To Reproduce

```python
from fastcs.controller import Controller
from fastcs.attributes import AttrRW
from fastcs.datatypes import Int
from fastcs.launch import FastCS
from fastcs.transport.epics.options import EpicsIOCOptions, EpicsOptions

class TestValueConroller(Controller):
    value = AttrRW(Int())


if __name__ == "__main__":
    epics_options = EpicsOptions(ioc=EpicsIOCOptions(pv_prefix="DEVICE"))
    controller = TestValueConroller()
    fastcs = FastCS(controller, epics_options)
    fastcs.run()

```

The PV works as expected:

```bash
─>$ caput DEVICE:Value 10000
Old : DEVICE:Value                   0
New : DEVICE:Value                   10000
...
─>$ caget DEVICE:Value
DEVICE:Value                   10000
```

but in the IOC console:

```python
>>> controller.value._value
0
```

regardless of what the PV is set to

GDYendell:
This allows creating Attributes without a Handler for parameters that are only used for internal controller logic and not sent directly to a device.

Fixes #101

### DIFF
--- a/src/fastcs/attributes.py
+++ b/src/fastcs/attributes.py
@@ -41,6 +41,19 @@ class Handler(Sender, Updater, Protocol):
     pass
 
 
+class SimpleHandler(Handler):
+    """Handler for internal parameters"""
+
+    async def put(self, controller: Any, attr: AttrW, value: Any):
+        await attr.update_display_without_process(value)
+
+        if isinstance(attr, AttrRW):
+            await attr.set(value)
+
+    async def update(self, controller: Any, attr: AttrR):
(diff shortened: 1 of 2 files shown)


dperl-dls (approved):
This is great, exactly what I needed, thank you!

PR #108, opened 2024-12-19
add unpack_status_array method

LuisFSegalla:
Add option to unpack an array of parameters creating a PV for each value of the array instead of presenting it as a list.

### DIFF
--- a/src/fastcs/util.py
+++ b/src/fastcs/util.py
@@ -3,3 +3,42 @@ def snake_to_pascal(input: str) -> str:
     return "".join(
         part.title() if part.islower() else part for part in input.split("_")
     )
+
+def unpack_status_arrays(parameter: list, uri: list[list[str]]):
+    """Takes a list of parameters and a list of special uri. Search the parameter
+    for elements that match the values in the uri list and split them into one
+    new odinParameter for each value.
+
+    Args:
+        parameter: List of parameters
+        uri: List of special uris to search and replace
+
+    Returns:
+        list[parameters]
+
+    """
+    removelist = []
+    for el in parameter:
+        if el.uri in uri:
+            status_list = (
+                el.metadata["value"]
+                .replace(",", "")
+                .replace("'", "")
+                .replace("[", "")
+                .replace("]", "")
+                .split()
+            )
+            for idx, value in enumerate(status_list):
+                metadata = {
+                    "value": value,
+                    "type": el.metadata["type"],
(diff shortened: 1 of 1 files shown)


src/fastcs/util.py
```diff
@@ -3,3 +3,42 @@ def snake_to_pascal(input: str) -> str:
     return "".join(
         part.title() if part.islower() else part for part in input.split("_")
     )
+
+def unpack_status_arrays(parameter: list, uri: list[list[str]]):
```
GDYendell, line 7:
```suggestion
def unpack_status_arrays(parameters: list[OdinParameter], uri: list[list[str]]):
```

src/fastcs/util.py
```diff
@@ -3,3 +3,42 @@ def snake_to_pascal(input: str) -> str:
...
+    Args:
+        parameter: List of parameters
+        uri: List of special uris to search and replace
+
+    Returns:
+        list[parameters]
+
+    """
+    removelist = []
+    for el in parameter:
```
GDYendell, line 21:
```suggestion
    for parameter in parameters:
```

src/fastcs/util.py
```diff
@@ -3,3 +3,42 @@ def snake_to_pascal(input: str) -> str:
...
+        if el.uri in uri:
+            status_list = (
+                el.metadata["value"]
+                .replace(",", "")
+                .replace("'", "")
+                .replace("[", "")
+                .replace("]", "")
+                .split()
+            )
```
GDYendell, line 30: This is a bit horrible and I don't think it is necessary because it will populate the value on the initial poll. The key thing is to make the separate parameters with the index on the end of the URI. Could you try setting the value to `""` and then checking that it is updated once the poll loop starts?

Then you also construct the metadata dict once and use it for all the new parameters.

GDYendell (commented):
First pass

src/fastcs/util.py
LuisFSegalla: Wouldn't make a list of **OdinParameter** be kind of a problem? Because OdinParameter is a class inside fastcs-odin which depends on fastCS

GDYendell: Oh! This is a good point. I think this needs to work in a completely different way, but I think it will be a bit more work. I will make an issue for this.

Let's go back to the original fastcs-odin PR for now, sorry!

PR #113, opened 2025-01-22
Coro sleep timing

marcelldls:
Sleeping after the scan coroutines means that in principle this is a constant delay rather than a fixed period


marcelldls:
Related to #68

GDYendell (approved):

PR #115, opened 2025-01-24
Add get_controller_schema function

marcelldls:
This adds a `get_controller_schema` function that can be convenient for assessing the schema of a controller - useful for the testing of a FastCS project. Another benefit is that the `tests/data/schema.json` is now also assessed by the new test - so the test will fail if its not kept up to date

### DIFF
--- a/tests/data/schema.json
+++ b/tests/data/schema.json
@@ -3,7 +3,7 @@
     "EpicsDocsOptions": {
       "properties": {
         "path": {
-          "default": "/home/esq51579/WIP/FastCS",
+          "default": ".",
           "format": "path",
           "title": "Path",
           "type": "string"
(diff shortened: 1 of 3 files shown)


tests/data/schema.json
```diff
@@ -35,7 +35,7 @@
     "EpicsGUIOptions": {
       "properties": {
         "output_path": {
-          "default": "/home/esq51579/WIP/FastCS/output.bob",
+          "default": "output.bob",
```
GDYendell, line 38: I guess this is due to #110 ?

GDYendell (dismissed):

tests/data/schema.json
marcelldls: Yes - I failed to update the schema. But now we have a test that will fail

GDYendell (approved):

PR #128, opened 2025-03-10
fix: Allowed multiple update callbacks on an `AttrR`

evalott100:
* Transports will add an update callback, so if we don't allow for multiple then running them simultaneously won't work.
* The controller level can add their own now too.

### DIFF
--- a/src/fastcs/attributes.py
+++ b/src/fastcs/attributes.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import asyncio
 from collections.abc import Callable
 from enum import Enum
 from typing import Any, Generic, Protocol, runtime_checkable
@@ -135,7 +136,7 @@ class AttrR(Attribute[T]):
         self._value: T = (
             datatype.initial_value if initial_value is None else initial_value
         )
-        self._update_callback: AttrCallback[T] | None = None
+        self._update_callbacks: list[AttrCallback[T]] | None = None
         self._updater = handler
 
     def get(self) -> T:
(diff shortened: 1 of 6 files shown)


jsouter:
Looks good, I'll check if with fastcs-eiger

src/fastcs/attributes.py
```diff
@@ -186,21 +189,26 @@ async def process(self, value: T) -> None:
...
 
     def has_process_callback(self) -> bool:
-        return self._process_callback is not None
+        return self._process_callbacks is not None and len(self._process_callbacks) != 0
 
-    def set_write_display_callback(self, callback: AttrCallback[T] | None) -> None:
-        self._write_display_callback = callback
+    def add_write_display_callback(self, callback: AttrCallback[T]) -> None:
+        if self._write_display_callbacks is None:
+            self._write_display_callbacks = []
```
evalott100, line 210: We have to do this because there is a problem with initialising `= []` in  `__init__` directly for class level attributes which are copied (the callback lists don't seem to also be copied).

src/fastcs/attributes.py
```diff
@@ -186,21 +189,26 @@ async def process(self, value: T) -> None:
...
     async def process_without_display_update(self, value: T) -> None:
-        if self._process_callback is not None:
-            await self._process_callback(self._datatype.validate(value))
+        if self._process_callbacks:
+            await asyncio.gather(*[cb(value) for cb in self._process_callbacks])
 
     async def update_display_without_process(self, value: T) -> None:
-        if self._write_display_callback is not None:
-            await self._write_display_callback(self._datatype.validate(value))
+        self._datatype.validate(value)
```
GDYendell, line 196: `validate` does type coercion of the value, so this still needs to be `value = ...validate(value)`

src/fastcs/attributes.py
```diff
@@ -186,21 +189,26 @@ async def process(self, value: T) -> None:
...
-    def set_process_callback(self, callback: AttrCallback[T] | None) -> None:
-        self._process_callback = callback
+    def add_process_callback(self, callback: AttrCallback[T]) -> None:
+        if self._process_callbacks is None:
+            self._process_callbacks = []
+        self._process_callbacks.append(callback)
 
     def has_process_callback(self) -> bool:
-        return self._process_callback is not None
+        return self._process_callbacks is not None and len(self._process_callbacks) != 0
```
GDYendell, line 206:
```suggestion
        return self._process_callbacks is not None and self._process_callbacks
```
empty lists are falsy

GDYendell (changes requested):

GDYendell (approved):

PR #134, opened 2025-03-11
Ensure backend creates sender callbacks with the correct controller

### LINKED ISSUE #130: `Sender`/`Updater`/`Handler` now incorrectly take the root controller
`Sender`, `Updater`, and `Handler` is now broken for subcontrollers. The `controller` passed as an arg to `update` and `put` is always the main `Controller` passed into FastCS, rather than the controller the attribute is registered under.

GDYendell:
Fixes #130

This is not the ideal fix, but tried many different things and they were all quite horrible:

- Adding the controller back into `ControllerAPI` - the transport should not have access to it
- Adding the `ScanCallback`s to `ControllerAPI` - this just doesn't really make sense, only the Backend needs them
- Make the `Backend` manually parse methods and attribute out of the `Controller` and only create the `ControllerAPI` at the end - this was OK, but meant repeating [this](https://github.com/DiamondLightSource/FastCS/blob/main/src/fastcs/backend.py#L171-L184) code multiple times.

 I think the interface of the handlers may not be correct, which is making this painful. It doesn't seem necessary for the whole controller to be passed into the handler put/update methods.

I don't know why that tango test is suddenly failing on 3.12 :cry: It doesn't happen on main or locally.

### DIFF
--- a/src/fastcs/backend.py
+++ b/src/fastcs/backend.py
@@ -30,7 +30,10 @@ class Backend:
     def _link_process_tasks(self):
         for controller_api in self.controller_api.walk_api():
             _link_put_tasks(controller_api)
-            _link_attribute_sender_class(controller_api, self._controller)
+            _link_attribute_sender_class(
+                controller_api,
+                self._controller.get_controller_by_path(controller_api.path),
+            )
 
     def __del__(self):
         self._stop_scan_tasks()
@@ -74,7 +77,7 @@ def _link_put_tasks(controller_api: ControllerAPI) -> None:
 
 
 def _link_attribute_sender_class(
-    controller_api: ControllerAPI, controller: Controller
+    controller_api: ControllerAPI, controller: BaseController
 ) -> None:
     for attr_name, attribute in controller_api.attributes.items():
         match attribute:
--- a/src/fastcs/controller.py
+++ b/src/fastcs/controller.py
@@ -4,6 +4,7 @@ from copy import copy
 from typing import get_type_hints
 
 from fastcs.attributes import Attribute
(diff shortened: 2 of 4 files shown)


evalott100:
> It doesn't seem necessary for the whole controller to be passed into the handler put/update methods.

I agree. I think that if a controller is necessary for the `Handler` to function then it should be an instance member on the `Handler`.

evalott100:
> I don't know why that tango test is suddenly failing on 3.12 😢 It doesn't happen on main or locally.

Feel free to cherry pick [this](https://github.com/DiamondLightSource/FastCS/commit/c8d2c6fa50691d91fd4196cf8cb3b08337a3032c).

I remember having this problem in ophyd-async too, I'm looking through pytango issues now.

jsouter:
`EigerHandler.update` is still receiving the top controller in fastcs-eiger with this fix, I think you also need to update `_get_scan_coros` to use the correct controller.

jsouter (changes requested):
Something like
```
def _get_scan_coros(
    root_controller_api: ControllerAPI, root_controller: Controller
) -> list[Callable]:
    scan_dict: dict[float, list[Callable]] = defaultdict(list)

    for controller_api in root_controller_api.walk_api():
        controller = root.controller.get_subcontroller_from_path(controller_api.path)
        _add_scan_method_tasks(scan_dict, controller_api)
        _add_attribute_updater_tasks(scan_dict, controller_api, controller)

    scan_coros = _get_periodic_scan_coros(scan_dict)
    return scan_coros

```

GDYendell:
Closing in preference for #135

PR #135, opened 2025-03-12
Add Handler.initialise(controller)

### LINKED ISSUE #59: Add Handler.initialise(controller) and call during Backend initialisation
This will enable handlers to do some logic based on fully initialised controllers that currently needs to be done in `update` every time. For example, filtering a list of child attributes to create status from. By default `initialise` will do nothing.

This will be called in `Controller.initialise` (which should be moved to `BaseController` as it will be needed in sub controllers too). Controllers that already implement `initialise` (fastcs-odin) will then have to call `super().initialise()`. It will need to be called with the controller and the attribute it is attached to. This is intended to enable https://github.com/DiamondLightSource/fastcs-odin/issues/21, so it probably makes sense to work on both issues at the same time to make sure the interface works.

## Acceptance Criteria
- Handlers can implement an initialise method that will be called with a initialised controller instance before the backend enters its run loop.

### LINKED ISSUE #130: `Sender`/`Updater`/`Handler` now incorrectly take the root controller

ajgdls:
Resolves #59 
Added default initialise methods to Sender and Updater base classes.
Added initialise method to Attribute, checks for handler and calls the method.
Moved initialise into BaseController class, loops over attributes calling initialise with a reference to self.
Added unit test for Attribute class.

Fixes #59 
Fixes #130

### DIFF
--- a/src/fastcs/attributes.py
+++ b/src/fastcs/attributes.py
@@ -22,16 +22,20 @@ class AttrMode(Enum):
 class Sender(Protocol):
     """Protocol for setting the value of an ``Attribute``."""
 
-<<<<<<< HEAD
-    async def put(
-        self, controller: fastcs.controller.BaseController, attr: AttrW, value: Any
-    ) -> None:
-=======
-    async def initialise(self, controller: Any):
+    # Record the controller that owns this attribute handler
+    controller: Any = None
+
+    async def initialise(self, controller: Any) -> None:
+        # Register the controller
+        self.controller = controller
+
+        # Continue with initialisation
+        await self._initialise()
+
+    async def _initialise(self) -> None:
         pass
 
-    async def put(self, controller: Any, attr: AttrW, value: Any) -> None:
->>>>>>> 43721ff (Added default initialise methods to Sender and Updater base classes. Added initialise method to Attribute, checks for handler and calls the method. Moved initialise into BaseController class, loops over attributes calling initialise with a reference to self. Added unit test for Attribute class.)
+    async def put(self, attr: AttrW, value: Any) -> None:
         pass
 
 
@@ -39,19 +43,25 @@ class Sender(Protocol):
 class Updater(Protocol):
     """Protocol for updating the cached readback value of an ``Attribute``."""
 
+    # Record the controller that owns this attribute handler
+    controller: Any = None
+
     # If update period is None then the attribute will not be updated as a task.
     update_period: float | None = None
 
-<<<<<<< HEAD
-    async def update(
-        self, controller: fastcs.controller.BaseController, attr: AttrR
-    ) -> None:
-=======
-    async def initialise(self, controller: Any):
+    async def initialise(self, controller: Any) -> None:
         pass
 
-    async def update(self, controller: Any, attr: AttrR) -> None:
->>>>>>> 43721ff (Added default initialise methods to Sender and Updater base classes. Added initialise method to Attribute, checks for handler and calls the method. Moved initialise into BaseController class, loops over attributes calling initialise with a reference to self. Added unit test for Attribute class.)
+        # Register the controller
+        self.controller = controller
+
+        # Continue with initialisation
+        await self._initialise()
+
+    async def _initialise(self) -> None:
+        pass
+
+    async def update(self, attr: AttrR) -> None:
         pass
 
 
@@ -65,15 +75,13 @@ class Handler(Sender, Updater, Protocol):
 class SimpleHandler(Handler):
     """Handler for internal parameters"""
 
-    async def put(
-        self, controller: fastcs.controller.BaseController, attr: AttrW, value: Any
-    ):
+    async def put(self, attr: AttrW, value: Any):
         await attr.update_display_without_process(value)
 
         if isinstance(attr, AttrRW):
             await attr.set(value)
 
-    async def update(self, controller: Any, attr: AttrR):
+    async def update(self, attr: AttrR):
         raise RuntimeError("SimpleHandler cannot update")
 
 
--- a/src/fastcs/attributes.py
+++ b/src/fastcs/attributes.py
@@ -22,9 +22,11 @@ class AttrMode(Enum):
 class Sender(Protocol):
     """Protocol for setting the value of an ``Attribute``."""
 
-    async def initialise(self, controller: Any) -> None: ...
+    async def initialise(self, controller: fastcs.controller.BaseController) -> None:
+        pass
 
-    async def put(self, attr: AttrW[T], value: T) -> None: ...
+    async def put(self, attr: AttrW[T], value: T) -> None:
+        pass
 
 
 @runtime_checkable
@@ -34,9 +36,11 @@ class Updater(Protocol):
     # If update period is None then the attribute will not be updated as a task.
     update_period: float | None = None
 
-    async def initialise(self, controller: Any) -> None: ...
+    async def initialise(self, controller: fastcs.controller.BaseController) -> None:
+        pass
 
-    async def update(self, attr: AttrR) -> None: ...
+    async def update(self, attr: AttrR) -> None:
+        pass
 
 
 @runtime_checkable
@@ -48,9 +52,6 @@ class Handler(Sender, Updater, Protocol):
 
 class SimpleHandler(Handler):
     """Handler for internal parameters"""
-
-    async def initialise(self, controller: Any) -> None:
-        pass
 
     async def put(self, attr: AttrW[T], value: T) -> None:
         await attr.update_display_without_process(value)
@@ -107,7 +108,7 @@ class Attribute(Generic[T]):
     def group(self) -> str | None:
         return self._group
 
-    async def initialise(self, controller: Any) -> None:
+    async def initialise(self, controller: fastcs.controller.BaseController) -> None:
         if self._handler is not None:
             await self._handler.initialise(controller)
 
--- a/src/fastcs/attributes.py
+++ b/src/fastcs/attributes.py
@@ -36,7 +36,7 @@ class AttrHandlerR(_BaseAttrHandler):
     # If update period is None then the attribute will not be updated as a task.
     update_period: float | None = None
 
-    async def update(self, attr: AttrR) -> None:
+    async def update(self, attr: AttrR[T]) -> None:
         pass
 
 
--- a/src/fastcs/attributes.py
+++ b/src/fastcs/attributes.py
@@ -18,19 +18,19 @@ class AttrMode(Enum):
     READ_WRITE = 3
 
 
-class _BaseHandler:
+class _BaseAttrHandler:
     async def initialise(self, controller: fastcs.controller.BaseController) -> None:
         pass
 
 
-class Setter(_BaseHandler):
+class AttrHandlerW(_BaseAttrHandler):
     """Protocol for setting the value of an ``Attribute``."""
 
     async def put(self, attr: AttrW[T], value: T) -> None:
         pass
 
 
-class Updater(_BaseHandler):
+class AttrHandlerR(_BaseAttrHandler):
     """Protocol for updating the cached readback value of an ``Attribute``."""
 
     # If update period is None then the attribute will not be updated as a task.
@@ -40,13 +40,13 @@ class Updater(_BaseHandler):
         pass
 
 
-class Handler(Setter, Updater):
-    """Protocol encapsulating both ``Setter`` and ``Updater``."""
+class AttrHandlerRW(AttrHandlerR, AttrHandlerW):
+    """Protocol encapsulating both ``AttrHandlerR`` and ``AttHandlerW``."""
 
     pass
 
 
-class SimpleHandler(Handler):
+class SimpleAttrHandler(AttrHandlerRW):
     """Handler for internal parameters"""
 
     async def put(self, attr: AttrW[T], value: T) -> None:
@@ -131,7 +131,7 @@ class AttrR(Attribute[T]):
         datatype: DataType[T],
         access_mode=AttrMode.READ,
         group: str | None = None,
-        handler: Updater | None = None,
+        handler: AttrHandlerR | None = None,
         initial_value: T | None = None,
         description: str | None = None,
     ) -> None:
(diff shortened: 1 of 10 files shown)


src/fastcs/controller.py
```diff
@@ -27,6 +27,11 @@ def __init__(
 
         self._bind_attrs()
 
+    async def initialise(self) -> None:
+        # Loop over attributes and initialise any registered handlers
+        for attr in self.attributes.values():
+            await attr.initialise(self)
```
jsouter, line 33: Would this be better as an asyncio.gather?

jsouter:
Could be helpful for setting up things like converters so we can have less logic inside the update/put calls!

src/fastcs/controller.py
evalott100: Definitely would be, Handler.initialise could be doing io.

evalott100 (changes requested):
I think as part of this change we should take out `controller` as an argument to `put`/`update`, making [this PR](https://github.com/DiamondLightSource/FastCS/pull/134) unneccesary.

If the user wants the controller in the handler for use in `put` or `update` then they should store it as an attribute (in the python sense) in their handler from initialise.

We can take out controller as an argument here

https://github.com/DiamondLightSource/FastCS/blob/5aa7a1216739be340827ca0f24d2ce7bf3ed925c/src/fastcs/backend.py#L76-L87

and here

https://github.com/DiamondLightSource/FastCS/blob/5aa7a1216739be340827ca0f24d2ce7bf3ed925c/src/fastcs/backend.py#L130-L141

src/fastcs/controller.py
GDYendell: Agreed, we just need to make sure it is called in a context that makes sense.

The controller should be initialised to create any dynamic attributes and methods at this point and I don't think these methods should (although it is python so we can't stop it) modify the controller or attributes, so there will be no clashes between them running concurrently.

This should just be used for things like populating a list of attributes that the handler should poke when another is updated, so that it doesn't have to search for them every time, like [here](https://github.com/DiamondLightSource/fastcs-odin/blob/main/src/fastcs_odin/odin_adapter_controller.py#L81).

tests/test_attribute.py
```diff
@@ -60,6 +60,18 @@ async def test_simple_handler_rw(mocker: MockerFixture):
...
+@pytest.mark.asyncio
+async def test_handler_initialise(mocker: MockerFixture):
+    handler = SimpleHandler()
+    handler_mock = mocker.patch.object(handler, "initialise")
+    attr = AttrR(Int(), handler=handler)
+
+    await attr.initialise(mocker.ANY)
+
+    # The handler initialise method should be called from the attribute
+    handler_mock.assert_called_once_with(1)
```
GDYendell, line 72: Why `1`?

ajgdls: That's a mistake, I was thinking about it being called once!

src/fastcs/controller.py
ajgdls: Sure thing, agreed.

ajgdls: @GDYendell you *could* split this part of the initialise away (by naming it something else and leaving the original initialise) from the initialise which is overidden, therefore removing the need to have this explicitly called in an overridden method.  It's an option which then removes the ability for a developer to not know they should have called `super`

src/fastcs/attributes.py
```diff
@@ -19,24 +19,46 @@ class AttrMode(Enum):
...
         pass
 
 
 @runtime_checkable
 class Updater(Protocol):
     """Protocol for updating the cached readback value of an ``Attribute``."""
 
+    # Record the controller that owns this attribute handler
+    controller: Any = None
+
```
evalott100, line 45:
``` suggestion
```

I don't think we need to put controller as a field on every `Updater`/`Sender`/`Handler`.

I'd be happy for `Handler.initialise` to take it and if downstream wants one in their implementation then they can

``` python
class MyHandler(Handler):
    _my_controller: MyController

    def __init__(self):
        ...
    async def initialise(self, controller: MyController) -> None:
        self._controller = controller

    def update(self, attr: AttrR) -> None:
        await self._controller.do_something_now()
```

src/fastcs/attributes.py
```diff
@@ -19,24 +19,46 @@ class AttrMode(Enum):
 class Sender(Protocol):
     """Protocol for setting the value of an ``Attribute``."""
 
-    async def initialise(self, controller: Any):
+    # Record the controller that owns this attribute handler
+    controller: Any = None
+
```
evalott100, line 24:
``` suggestion
```

src/fastcs/attributes.py
```diff
@@ -19,24 +19,46 @@ class AttrMode(Enum):
...
+    async def initialise(self, controller: Any) -> None:
+        # Register the controller
+        self.controller = controller
+
+        # Continue with initialisation
+        await self._initialise()
+
+    async def _initialise(self) -> None:
         pass
```
evalott100, line 33:
``` suggestion
    async def intitialise(self, controller: Any) -> None:
        """Subclasses can implement their own method using the controller."""
```

src/fastcs/attributes.py
```diff
@@ -19,24 +19,46 @@ class AttrMode(Enum):
...
+        pass
+
+        # Register the controller
+        self.controller = controller
+
+        # Continue with initialisation
+        await self._initialise()
+
+    async def _initialise(self) -> None:
         pass
```
evalott100, line 59:
``` suggestion
    async def intitialise(self, controller: Any) -> None:
        """Subclasses can implement their own method using the controller."""
```

src/fastcs/attributes.py
ajgdls: Do we have concrete examples where the subclass does not require access to the controller?  I can take a look at the odin data fastcs module but I'm guessing there are several other fastcs modules I'm not aware of.

evalott100: Hmmm other than fastcs-pandablocks I don't think there's too many, and fastcs-pandablocks will probably change to using the controller in the handlers if we change `Connection`.

I'm still not that keen to have `Any` as a class type hint on `Handler` and force the user to implement

``` python
class MyHandler(Handler):
    controller: MyController

    async def initialise(self, controller: MyController) -> None:
        await super().initialise(controller)
```

to get a typed controller attribute, or alternatively

``` python
class MyHandler(Handler):
    async def initialise(self, controller: MyController) -> None:
        await super().initialise(controller)
```

for an untyped one.

I think it's nicer to have the subclasses type hint exactly which `Controller` they're expecting and just set it at that point if it's needed, especially if all `super().initialise()` does is `self.controller = controller`.

# Alternatively

We could alternatively use a `TypeVar`


``` python
ControllerT = TypeVar("ControllerT") # Would be nice if we could bind to `BaseController`
class Handler(Generic[ControllerT])
    controller: ControllerT
    async def initialise(self, controller: ControllerT):
        self.controller = controller
```

Then

``` python
class MyController(Handler[MyController])
```

would have a typed `my_controller.controller` if they call `super().initialise(controller)`, or if they don't implement `MyController.initialise()`

evalott100: > but doesn't that cause even more problems if you have cases where your doesn't need a controller?

Yes, but if there's going to be a `controller` attribute on every `Handler` I'd prefer for it to be typed.

ajgdls: It seems like there are a couple of different points to address here:
1. Do you want to remove the controller registration handling from the developer completely?
2. Do you want to ensure typed controllers?

From my ancient OOP ideals I always like the idea of removing the need for a developer to do something if it can be abstracted and them presented with a simple interface.

However, I'm definitely not an expert on FastCS and I haven't experience on developing FastCS Handlers or controllers, and so I don't mind implementing the solutions suggested, I just need a definitive "do it this way".  Who gets the final say @GDYendell @evalott100 ?

GDYendell: Given @evalott100 is leaving I guess me, but the problem is we disagree. I think the problem is we need to flesh out this idea some more to have a clear idea of what we want.

Another complication to throw into the mix: Handlers are currently just Protocols - they are not inherited. So, if we want some base functionality to save driver developers registering the controller themselves that would also have to change, which would break all current drivers.

ajgdls: Yes I agree that makes sense to have a good idea of your end goals first before committing to an implementation that might trip you up later on.

GDYendell: Let's do this:

```python
class Sender(Protocol):

    async def initialise(self, controller: BaseController):
        ...

    async def put(attr: Attribute[T], value[T]):
        ...
```

Then we can do

```python
class MyController(Controller):
    ...

    async def specific_send_method(self, attr: AttrW[T], value: T):
        ...

class MySender():

    controller: MyController | None = None

    async def initialise(self, controller: MyController):  # This might give incompatible override? In which case in can do an assert isinstance MyController
        self.controller = controller
    
    async def put(attr: AttrW[T], value: T):
        if controller is None:
            raise

        self.controller.specific_send_method(attr, value)
```

We do have an example where the handler doesn't need the controller - SimpleHandler.

GDYendell: This is a reason to swap from `Protocol` to inheritance. `SimpleHandler` will have to implement `initialise` to fulfil the protocol even though it doesn't have anything to do - with inheritance it can just call the base method.

evalott100: Definitely agree inheritence, I don't think there's a good reason for these to be protocols.

tests/test_attribute.py
```diff
@@ -53,13 +53,41 @@ async def test_simple_handler_rw(mocker: MockerFixture):
...
+@pytest.mark.asyncio
+async def test_handler_initialise(mocker: MockerFixture):
+    handler = SimpleHandler()
+    handler_mock = mocker.patch.object(handler, "initialise")
+    attr = AttrR(Int(), handler=handler)
+
+    await attr.initialise(mocker.ANY)
+
+    # The handler initialise method should be called from the attribute
+    handler_mock.assert_called_once_with(mocker.ANY)
```
GDYendell, line 76: This should make a named `MagicMock` to pass to `attr.initialise` and then check the `handler_mock` is called with that.
For assertions where you only care about some of the arguments you can use mocks to check those and fill the rest with `mocker.ANY`.

src/fastcs/controller.py
```diff
@@ -29,6 +30,14 @@ def __init__(
 
         self._bind_attrs()
 
+    async def initialise(self) -> None:
```
GDYendell, line 33: I don't like that we will always have to call super if this is overridden in a base class. It isn't really an override because it is doing a different thing - it is an implementation detail that should be hidden from the public API.

It would be nicer if we keep

```python
async def initialise(self):
    pass
```

to be overridden if needed, put this new logic in an `initialise_attributes` method and just have the backend call both

```python
controller.initialise()
controller.initialise_attributes()
``` 

What do you think?

ajgdls: Yes I agree this would improve the public API.

ajgdls: I've pushed the change as I think you would like it @GDYendell

src/fastcs/attributes.py
```diff
@@ -22,9 +22,10 @@ class AttrMode(Enum):
 class Sender(Protocol):
```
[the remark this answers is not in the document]
ajgdls: @GDYendell has previously mentioned that, but I've slightly lost track of why they wouldn't now be Protocol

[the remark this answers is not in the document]
ajgdls: I think specifically we would have a Handler inheriting from a Sender and an Updater where both of those have initialise methods with the same signature.  I don't think that is very elegant.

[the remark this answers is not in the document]
GDYendell: What if we have a `_BaseHandler` that implements `initialise`, `Updater` and `Sender` both inherit that, and `Handler` inherits both? I realise this makes a diamond, but I don't think that is a problem in this case.

Also, I am open to alternative names for `Handler` as I don't think it is very confusing. Now would be a good time to change it.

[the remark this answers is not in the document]
ajgdls: Yeah sorry I missed your previous comment.  A _BaseHandler sounds OK to me, I prefer the idea of a single implementation at the top of the Diamond.  For the name, I actually wonder if Sender should be Setter?  Sender makes it sound like some sort of communication class.  I'm not sure I can come up with an improved alternative for Handler though.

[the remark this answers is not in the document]
GDYendell: I think it would make sense to mirror the R/W/RW naming convention for this. So we could have `AttrHandler`, `AttrHandlerR`, `AttrHandlerW`, `AttrHandlerRW`. They would still have `set` and `update` methods.

This also makes it completely obvious what that if you have an `AttrRW` you need to give it an `AttrHandlerRW`, not just an `AttrHandlerR`, whereas currently you need to know to pass it a `Handler` instead of an `Updater`.

I still don't particularly like the word Handler, but these names add some context that make it clearer what it means. Could also be `AttrLink`, `AttrConnector`, `AttrChannel`? Any other options you can think of?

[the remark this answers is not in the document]
ajgdls: @GDYendell  I have pushed the name change (because they had already been changed and so I think it makes sense to get it done.

GDYendell:
This is actually broken and is showing up a lack of testing. The updater callbacks are still being called with a controller [here](https://github.com/DiamondLightSource/FastCS/blob/main/src/fastcs/backend.py#L135), which causes them to stop and print a warning.

The reason this exception is caught and printed was originally so that one attribute update failing doesn't crash the whole application. I was on the fence about this and am again thinking just crashing would be better.

ajgdls:
Good spot, I guess you noticed this by running an application?  Perhaps I should add the type hints as this is not really a runtime exception but a develop time syntax error.

PR #141, opened 2025-03-20
Add Blocking Flag to Command PVs

shihab-dls:
When we put to a command PV, the put returns immediately, not waiting for the bound callback to complete. Adding the `blocking=True` flag sets the PVs `PACT` field high until the bound callback sets the PVs status to completed. This should allow an ophyd-async `SignalX` to be properly awaited.

I'm not sure if this behaviour should be set like this, however; is it always the case we'd want to wait for a callback when processing a command PV?

### DIFF
--- a/src/fastcs/transport/epics/pva/_pv_handlers.py
+++ b/src/fastcs/transport/epics/pva/_pv_handlers.py
@@ -83,9 +83,9 @@ class CommandPvHandler:
 
             # Flip to true once command task starts
             pv.post({"value": True, **p4p_timestamp_now(), **p4p_alarm_states()})
-            op.done()
             alarm_states = await self._run_command()
             pv.post({"value": False, **p4p_timestamp_now(), **alarm_states})
+            op.done()
         else:
             raise RuntimeError("Commands should only take the value `True`.")
 
--- a/src/fastcs/transport/epics/pva/_pv_handlers.py
+++ b/src/fastcs/transport/epics/pva/_pv_handlers.py
@@ -7,7 +7,7 @@ from p4p.server import ServerOperation
 from p4p.server.asyncio import SharedPV
 
 from fastcs.attributes import Attribute, AttrR, AttrRW, AttrW
-from fastcs.cs_methods import CommandCallback
+from fastcs.cs_methods import CommandCallback, CommandMode
 from fastcs.datatypes import Table
 
 from .types import (
@@ -51,8 +51,9 @@ class WritePvHandler:
 
 
 class CommandPvHandler:
-    def __init__(self, command: CommandCallback):
+    def __init__(self, command: CommandCallback, mode: CommandMode):
         self._command = command
+        self._mode = mode
         self._task_in_progress = False
 
     async def _run_command(self) -> dict:
@@ -81,11 +82,13 @@ class CommandPvHandler:
                     "Maybe the command should spawn an asyncio task?"
                 )
 
-            # Flip to true once command task starts
             pv.post({"value": True, **p4p_timestamp_now(), **p4p_alarm_states()})
-            op.done()
+            if self._mode == CommandMode.HIGH_AFTER_START:
+                op.done()
             alarm_states = await self._run_command()
             pv.post({"value": False, **p4p_timestamp_now(), **alarm_states})
+            if self._mode == CommandMode.HIGH_AFTER_FINISH:
+                op.done()
         else:
             raise RuntimeError("Commands should only take the value `True`.")
 
@@ -123,7 +126,7 @@ def make_shared_pv(attribute: Attribute) -> SharedPV:
     return shared_pv
 
 
-def make_command_pv(command: CommandCallback) -> SharedPV:
+def make_command_pv(command: CommandCallback, mode: CommandMode) -> SharedPV:
     type_ = NTScalar.buildType("?", display=True, control=True)
 
     initial = Value(type_, {"value": False, **p4p_alarm_states()})
@@ -133,7 +136,7 @@ def make_command_pv(command: CommandCallback) -> SharedPV:
 
     shared_pv = SharedPV(
         initial=initial,
-        handler=CommandPvHandler(command),
+        handler=CommandPvHandler(command, mode),
         wrap=_wrap,
     )
 
--- a/src/fastcs/transport/epics/pva/_pv_handlers.py
+++ b/src/fastcs/transport/epics/pva/_pv_handlers.py
@@ -73,6 +73,7 @@ class CommandPvHandler:
     async def put(self, pv: SharedPV, op: ServerOperation):
         value = op.value()
         raw_value = value["value"]
+        blocking = False
 
         if raw_value is True:
             if self._task_in_progress:
@@ -81,11 +82,20 @@ class CommandPvHandler:
                     "Maybe the command should spawn an asyncio task?"
                 )
 
+            # Try to retrieve block option from record field
+            requests = op.pvRequest()
+            if requests:
+                record_options = requests.get("record", {}).get("_options", {})
+                blocking = record_options.get("block") == "true"
+
             # Flip to true once command task starts
             pv.post({"value": True, **p4p_timestamp_now(), **p4p_alarm_states()})
+            if not blocking:
+                op.done()
             alarm_states = await self._run_command()
             pv.post({"value": False, **p4p_timestamp_now(), **alarm_states})
-            op.done()
+            if blocking:
+                op.done()
         else:
             raise RuntimeError("Commands should only take the value `True`.")
 
--- a/src/fastcs/transport/epics/pva/_pv_handlers.py
+++ b/src/fastcs/transport/epics/pva/_pv_handlers.py
@@ -73,7 +73,6 @@ class CommandPvHandler:
     async def put(self, pv: SharedPV, op: ServerOperation):
         value = op.value()
         raw_value = value["value"]
-        blocking = False
 
         if raw_value is True:
             if self._task_in_progress:
@@ -82,11 +81,12 @@ class CommandPvHandler:
                     "Maybe the command should spawn an asyncio task?"
                 )
 
-            # Try to retrieve block option from record field
-            requests = op.pvRequest()
-            if requests:
-                record_options = requests.get("record", {}).get("_options", {})
-                blocking = record_options.get("block") == "true"
+            # Check if record block request recieved
+            match op.pvRequest().todict():
+                case {"record": {"_options": {"block": "true"}}}:
+                    blocking = True
+                case _:
+                    blocking = False
 
             # Flip to true once command task starts
             pv.post({"value": True, **p4p_timestamp_now(), **p4p_alarm_states()})
--- a/src/fastcs/transport/epics/pva/_pv_handlers.py
+++ b/src/fastcs/transport/epics/pva/_pv_handlers.py
@@ -7,7 +7,7 @@ from p4p.server import ServerOperation
 from p4p.server.asyncio import SharedPV
 
 from fastcs.attributes import Attribute, AttrR, AttrRW, AttrW
-from fastcs.cs_methods import CommandCallback, CommandMode
+from fastcs.cs_methods import CommandCallback
 from fastcs.datatypes import Table
 
 from .types import (
@@ -51,9 +51,8 @@ class WritePvHandler:
 
 
 class CommandPvHandler:
-    def __init__(self, command: CommandCallback, mode: CommandMode):
+    def __init__(self, command: CommandCallback):
         self._command = command
-        self._mode = mode
         self._task_in_progress = False
 
     async def _run_command(self) -> dict:
@@ -82,13 +81,11 @@ class CommandPvHandler:
                     "Maybe the command should spawn an asyncio task?"
                 )
 
+            # Flip to true once command task starts
             pv.post({"value": True, **p4p_timestamp_now(), **p4p_alarm_states()})
-            if self._mode == CommandMode.HIGH_AFTER_START:
-                op.done()
+            op.done()
             alarm_states = await self._run_command()
             pv.post({"value": False, **p4p_timestamp_now(), **alarm_states})
-            if self._mode == CommandMode.HIGH_AFTER_FINISH:
-                op.done()
         else:
             raise RuntimeError("Commands should only take the value `True`.")
 
@@ -126,7 +123,7 @@ def make_shared_pv(attribute: Attribute) -> SharedPV:
     return shared_pv
 
 
-def make_command_pv(command: CommandCallback, mode: CommandMode) -> SharedPV:
+def make_command_pv(command: CommandCallback) -> SharedPV:
     type_ = NTScalar.buildType("?", display=True, control=True)
 
     initial = Value(type_, {"value": False, **p4p_alarm_states()})
@@ -136,7 +133,7 @@ def make_command_pv(command: CommandCallback, mode: CommandMode) -> SharedPV:
 
     shared_pv = SharedPV(
         initial=initial,
-        handler=CommandPvHandler(command, mode),
+        handler=CommandPvHandler(command),
         wrap=_wrap,
     )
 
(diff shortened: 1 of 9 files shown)


evalott100:
> I'm not sure if this behaviour should be set like this, however; is it always the case we'd want to wait for a callback when processing a command PV?

Hmmm this may require some thought on how to standardise.

# PVA

In pva I set the command widget to true while the command is running, then false after it has finished.

https://github.com/DiamondLightSource/FastCS/blob/1cce8ea7cbe0e7b31c9f294891f80bd8bbc60079/src/fastcs/transport/epics/pva/_pv_handlers.py#L53-L90

This means that `awaiting` a put to a `SignalX` will wait for the command to start runnning rather than for the command to finish.

I don't mind changing the standard to be that we do blocking until the command is complete, but if so we should agree on this standard.

Unless I'm mistaken, the pva analogue of the above would be to

``` python
             # Flip to true after the command ends
             alarm_states = await self._run_command()
             pv.post({"value": True, **p4p_timestamp_now(), **alarm_states}) 
             op.done()
```

and remove the `True`/`False` logic, a command records value is always `True`.

GDYendell:
I could imagine you might want either... so it probably needs to be a flag on `Command` and `@command`?

evalott100:
> I could imagine you might want either... so it probably needs to be a flag on `Command` and `@command`?

I was thinking the same, though we probably want it to be a kwarg of type:

```python
class CommandMode(enum.Enum):
    #: Command value becomes `True` in the transport only
    #: once the command starts, becomes `False` after it 
    #: finishes.
    CONTINUE_AFTER_START = "CONTINUE_ONCE_STARTED"

    #: Command value becomes `True` in the transport only after
    #: the command has finished. Becomes `False` immediately after.
    CONTINUE_AFTER_FINISH = "BLOCKING"
```

GDYendell:
I think `CONTINUE_AFTER_FINISH` should be high after (or just before) start, low after finish.

evalott100:
```python

class CommandMode(enum.Enum):
    #: Command value becomes `True` in the transport once
    #: the command starts, becomes `False` after it finishes.
    HIGH_AFTER_START = "HIGH_AFTER_START"

    #: Command value becomes `True` in the transport only after
    #: the command has finished. Becomes `False` immediately after.
    HIGH_AFTER_FINISH = "HIGH_AFTER_FINISH"
```

GDYendell:
@shihab-dls is there an issue to do with arming eiger that this would fix?

shihab-dls:
> @shihab-dls is there an issue to do with arming eiger that this would fix?

This should unblock [#551](https://github.com/bluesky/ophyd-async/issues/551), where the Eiger's `arm` signal is now a `SignalX`. This is incompatible with current functionality of ophyd-async's `set_and_wait_for_other_value`, which we used to use to wait for the detectors `State` signal to flag that arming is complete. So blocking on the pvs put here would avert the need to use this function at all.

GDYendell:
@coretl @DominicOram  do you have any thoughts on the behaviour we might want here?

shihab-dls:
I've just pushed some rough commits outlining the behaviour we're describing (to my understanding). @GDYendell and @evalott100 did I follow correctly? Happy to revert the changes if it's decided `HIGH_AFTER_FINISH` is all we need.

coretl:
I think we always want `blocking=True`, as we can always to the areaDetector thing of `caput -c` in the background and wait until the readback is True to tell an operation has started, and still have something to wait on to tell when it finishes.

So the order should be:
- Receive a `caput`/`pvput`
- Start the task
- Set the value to `True`
- Wait for the task to complete
- Set the value to `False`
- `pv.post`/finish the `on_update` call

GDYendell:
So we don't need the option @shihab-dls, it should always do Action(blocking=True) and the pva version should be updated to match @coretl's description above.

evalott100:
> So the order should be:
> 
>     * Receive a `caput`/`pvput`
> 
>     * Start the task
> 
>     * Set the value to `True`
> 
>     * Wait for the task to complete
> 
>     * Set the value to `False`
> 
>     * `pv.post`/finish the `on_update` call

This would be equivalent to what we have in the PVA transport right now, and the CA without setting the record to false again (I don't think this happens with a `builder.Action`).

# Suggestion

For equivalence to the p4p backend we change from using a `builder.Action` to a `builder.BoolOut` which runs an `on_update` which:
* On `True`: Runs the command and sets the record to `False`
* On `False`: Does nothing or raises an error.

shihab-dls:
> > blocking=True and blocking=False on builder.Action doesn't actually seem to make a difference in whether the caput on the pv times out, so we'll need to find out how to delay the operation to after the command ends.
> 
> I don't think this is what you found, is it @shihab-dls ?

I've been testing this out; it seems **it is true that the record value is set immediately**, which means that:

```python    
@command()
async def comm(self):
    print("Started command")
    await asyncio.sleep(5)
    print("Finished command")
```
with `blocking=True`, results in:

```bash
$ pvput -w 1 blah:Comm 1
Old : 2025-03-24 10:45:27.418  (0) 
New : 2025-03-24 10:48:52.454  (1) 
```
where the put does not timeout. Similarly, a pvget results in:

```bash
$ pvget -w 1 blah:Comm
blah:Comm 2025-03-24 10:50:21.889  (1)
```
where the new record value is immediately available.

However, the operation completion is only triggered **after** the callback completes, which mean monitors are updated **after the callback**:

```bash
$ pvput -w 1 blah:Comm 1
Old : 2025-03-24 11:03:32.414  (0) 
New : 2025-03-24 11:03:52.857  (1)
```

```bash
$ pvmonitor blah:Comm
blah:Comm 2025-03-24 11:03:32.414  (0) 
```
where the monitor does not update until 5 seconds later:
```bash
$ pvmonitor blah:Comm
blah:Comm 2025-03-24 11:03:32.414  (0) 
blah:Comm 2025-03-24 11:03:57.858  (1)
```
If you were to `pvget` the records `PACT` field instead of its value, you would see the required behaviour of:

> * Receive a `caput`/`pvput`
> * Start the task
> * Set the value to `True`
> * Wait for the task to complete
>* Set the value to `False`
>* `pv.post`/finish the `on_update` call

for example:

```
$ pvput -w 1 blah:Comm 1
Old : 2025-03-24 11:08:12.388  (0) 
New : 2025-03-24 11:08:28.699  (1) 
```

```
$ pvget blah:Comm.PACT
blah:Comm.PACT 2025-03-24 11:15:36.878  1 
```
followed by:
```
$ pvget blah:Comm.PACT
blah:Comm.PACT 2025-03-24 11:15:41.879  0 
```

which I believe is why @evvaaaa is concerned about us using 
>the flag instead of single value.

It seems, however, that a put with callback waits for server processing to be complete, which means that it correctly waits for the `PACT` field to go low before returning, giving us our desired `arm()` behaviour on detectors.

### TLDR
Setting `blocking=True` is not directly analogous to the changes in pva, because the records value field is updated immediately, rather than:
> * Set the value to `True`
> * Wait for the task to complete
>* Set the value to `False`

This means a `pvput` or `pvget` will return immediately.

However, the records `PACT` field does follow the order:
> * Set the value to `True`
> * Wait for the task to complete
>* Set the value to `False`

Which means that doing a put with callback will correctly wait for callback completion before returning.

DominicOram:
> @coretl @DominicOram do you have any thoughts on the behaviour we might want here?

I don't have much to contribute to the details of this but I agree with @coretl that the callback should only return on the process being completed. Otherwise we have no reliable way to tell something has completed from the client side. If a client chooses not to wait for that callback then that's up to individual clients

shihab-dls:
These final commits should now have this PR doing the following:

- Set `op.done()` after callback is complete for pva 
- Set `blocking=True` for ca 

which should have our put callbacks, regardless of backend, return **after** a bound method has completed.

coretl:
> These final commits should now have this PR doing the following:
> 
>     * Set `op.done()` after callback is complete for pva
> 
>     * Set `blocking=True` for ca
> 
> 
> which should have our put callbacks, regardless of backend, return **after** a bound method has completed.

Note that is for CA via the pythonSoftIOC transport, and PVA via the p4p transport. PVA via the pythonSoftIOC transport currently doesn't wait for completion due to https://github.com/epics-base/pvxs/issues/105. This is not a blocker for FastCS (as we will use PVA via the p4p transport, so will probably turn it off in pythonSoftIOC via https://github.com/DiamondLightSource/pythonSoftIOC/issues/185), but we will need this later for areaDetector IOCs when we go full PVA on beamlines.

GDYendell (approved):
I think this looks good now. I suggest we have a brief discussion in the stand up to make sure we are in agreement and then merge.

coretl:
We should make sure that we match the behaviour of the IOCs, so if we `pvput` it *doesn't* wait for completion, but if we `pvput -r 'record[block=true]' -w` then it does as per https://github.com/epics-base/pvxs/issues/105#issuecomment-2750956968

shihab-dls:
> We should make sure that we match the behaviour of the IOCs, so if we `pvput` it _doesn't_ wait for completion, but if we `pvput -r 'record[block=true]' -w` then it does as per [epics-base/pvxs#105 (comment)](https://github.com/epics-base/pvxs/issues/105#issuecomment-2750956968)

Looks like with `blocking=True` on the softIOC backend, `pvput -r 'record[block=true]' -w` does wait for the callback to complete.
```bash
$ time pvput -r "record[block=true]" -w 10 blah:Comm 1
Old : <undefined>              INVALID DRIVER UDF (0) 
New : 2025-03-26 12:29:02.427  (1) 

real	0m5.009s
```
where we timeout if `-w 4` (i.e., process takes > 4s)
```bash
$ time pvput -r "record[block=true]" -w 4 blah:Comm 1
Old : 2025-03-26 12:32:44.686  (0) 
Put timeout

real	0m4.009s
```
but the value still changes if we get afterwards:
```bash
$ pvget blah:Comm
blah:Comm 2025-03-26 12:33:02.479  (1)
```

### **Blocking Behaviour Post Merge:**

| Command  | SoftIOC WAITS? | P4P WAITS? |
| ------------- | ------------- | ------------- |
| `caput -c -w <TIMEOUT> <PREFIX>:<COMMAND_PV> <VALUE>`  | YES  | N/a |
| `caput -w <TIMEOUT> <PREFIX>:<COMMAND_PV> <VALUE>`  | NO | N/a  |
| `caput <PREFIX>:<COMMAND_PV> <VALUE>`  | NO | N/a  |
| `pvput -r 'record[block=true]' -w <TIMEOUT>  <PREFIX>:<COMMAND_PV> <VALUE>`  | YES | YES  |
| `pvput -w <TIMEOUT>  <PREFIX>:<COMMAND_PV> <VALUE>`  | NO | YES  |
| `pvput <PREFIX>:<COMMAND_PV> <VALUE>`  | NO | YES (with default timeout=5s)  |

Is this the behaviour that was agreed upon, @coretl and @GDYendell? Where **P4P** waits for callback completion **regardless** of additional flags (where -w is sort of required to ensure no timeout)?

src/fastcs/transport/epics/pva/_pv_handlers.py
```diff
@@ -83,9 +83,9 @@ async def put(self, pv: SharedPV, op: ServerOperation):
 
             # Flip to true once command task starts
             pv.post({"value": True, **p4p_timestamp_now(), **p4p_alarm_states()})
-            op.done()
```
coretl, line 86: This wants to be something like:
```python
blocking = op.pvRequest().something
if not blocking:
    op.done()
```
https://epics-base.github.io/p4p/server.html#p4p.server.ServerOperation.pvRequest

src/fastcs/transport/epics/pva/_pv_handlers.py
```diff
@@ -83,9 +83,9 @@ async def put(self, pv: SharedPV, op: ServerOperation):
...
             alarm_states = await self._run_command()
             pv.post({"value": False, **p4p_timestamp_now(), **alarm_states})
+            op.done()
```
coretl, line 88: Then this is
```suggestion
            if blocking:
                op.done()
```

src/fastcs/transport/epics/pva/_pv_handlers.py
```diff
@@ -81,11 +82,20 @@ async def put(self, pv: SharedPV, op: ServerOperation):
                     "Maybe the command should spawn an asyncio task?"
                 )
 
+            # Try to retrieve block option from record field
+            requests = op.pvRequest()
+            if requests:
+                record_options = requests.get("record", {}).get("_options", {})
+                blocking = record_options.get("block") == "true"
```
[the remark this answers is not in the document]
GDYendell: Whatever you end up doing could you add a comment saying it is what you get when the client does 

```
pvput -r "record[block=true]" ...
```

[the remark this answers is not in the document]
shihab-dls: This is way nicer; It's of a p4p `Value` type, but works if we `match op.pvRequest().todict()`.

shihab-dls:
This is now waiting on pythonSoftIOC [#186](https://github.com/DiamondLightSource/pythonSoftIOC/pull/186).

**Current state of this PR, with recent changes is:**
| Command  | SoftIOC WAITS? | P4P WAITS? |
| ------------- | ------------- | ------------- |
| `caput -c -w <TIMEOUT> <PREFIX>:<COMMAND_PV> <VALUE>`  | YES  | N/a |
| `caput -w <TIMEOUT> <PREFIX>:<COMMAND_PV> <VALUE>`  | NO | N/a  |
| `caput <PREFIX>:<COMMAND_PV> <VALUE>`  | NO | N/a  |
| `pvput -r 'record[block=true]' -w <TIMEOUT>  <PREFIX>:<COMMAND_PV> <VALUE>`  | N/a | YES  |
| `pvput -w <TIMEOUT>  <PREFIX>:<COMMAND_PV> <VALUE>`  | N/a | NO  |
| `pvput <PREFIX>:<COMMAND_PV> <VALUE>`  | N/a | NO  |

Where `PVA` is over `P4P`, `CA` is over `softIOC`, and waiting for callback completion matches [epics-base/pvxs#105 (comment)](https://github.com/epics-base/pvxs/issues/105#issuecomment-2750956968)

shihab-dls:
PR should be ready for re-review now.

**Note:**
We import `from pvi._format.dls import DLSFormatter`, which now fails linting with the new pyright release due to `reportPrivateImportUsage.` I've type ignored this for now, but it should be addressed in PVI at some point.

tests/transport/epics/pva/test_p4p.py
```diff
@@ -623,3 +627,45 @@ async def put_pvs():
...
+        )
+    except TimeoutError:
+        ...
+    serve.cancel()
+
+    assert len(command_runs_for_a_while_times) == 2
+
+    for i in range(2):
+        start, end = command_runs_for_a_while_times[i]
+        assert pytest.approx((end - start).total_seconds(), abs=0.05) == 0.2
```
GDYendell, line 671: Shouldn't the put with block=false return immediately?

shihab-dls: Yes, it does; probably not the most transparent test, but I use the same start time for both puts, and thus I'm measuring elapsed time. `block=True` takes `0.2s`, so the first appended time should be `0.2s`, then `block=False` takes no time, so the second appended time should also be `0.2s` (instead of `0.4s`).

A clearer test would be:
```python
for put_call, expected_duration in enumerate([0.2, 0]):
    start, end = command_runs_for_a_while_times[put_call]
    assert pytest.approx((end - start).total_seconds(), abs=0.05) == expected_duration
```
So, I'll make the change and re-request.

GDYendell: Thanks that makes much more sense to me!

GDYendell (approved):

tests/transport/epics/pva/test_p4p.py
```diff
@@ -623,3 +627,46 @@ async def put_pvs():
...
+
+    async def put_pvs():
+        ctxt = Context("pva")
+        for block in ["true", "false"]:
+            start_time = datetime.now()
+            await ctxt.put(
+                f"{pv_prefix}:CommandRunsForAWhile",
+                True,
+                request=f"record[block={block}]",
+            )
```
coretl, line 651: Looks like `ctxt.put(wait)` already does this:
https://github.com/bluesky/ophyd-async/pull/846

so we should be able to do:
```suggestion
        for block in [True, False]:
            start_time = datetime.now()
            await ctxt.put(
                f"{pv_prefix}:CommandRunsForAWhile",
                True,
                wait=block,
            )
```

coretl (approved):

GDYendell (approved):

Issue #149, opened 2025-04-02
Enum choices


jsouter:
Currently enum choice values are restricted to valid python names, as the choices are populated from the enum member names.
https://github.com/DiamondLightSource/FastCS/blob/2c546da833d390bf32806258acdc282c55c8b669/src/fastcs/transport/epics/ca/util.py#L71-L79

This restricts us from being able to have spaces in strings and other characters which would otherwise normally be allowed in ZRST, ONST, TWST,... in EPICS. I can understand why this is the case, but it would be nice if this could be supported. 

As a corollary,
see #150

GDYendell:
Are we happy that this can be worked around by constructing an Enum with a dict @jsouter ?

jsouter:
> Are we happy that this can be worked around by constructing an Enum with a dict [@jsouter](https://github.com/jsouter) ?

Yes, I think so

PR #151, opened 2025-04-03
CA enums with more than 16 values backed by string record

### LINKED ISSUE #150: CA Enums don't work properly with more than 16 choices
Splitting #149 into two issues
https://github.com/DiamondLightSource/FastCS/blob/2c546da833d390bf32806258acdc282c55c8b669/src/fastcs/transport/epics/ca/util.py#L85-L105

Enums with more than 16 values don't work correctly in Channel Access, I believe we just need to add extra logic to handle casting to/from epics types if `len(attribute.datatype.members) > MBB_MAX_CHOICES`. 
In Channel Access, `Enum`s with too many choices decay to `longin`/`longout`.

jsouter:
Closes #150

May need to add test? Note that I have tested this change against a WIP version of fastcs-eiger running against slightly reworked branch of FastCS #134.

Records with the Enum datatype whose enum_cls have more than 16 members now use `longStringIn`/`longStringOut` records instead of `longIn`/`longOut`, this allows us to still have labelled dropdowns and string readbacks

### DIFF
--- a/src/fastcs/transport/epics/ca/util.py
+++ b/src/fastcs/transport/epics/ca/util.py
@@ -87,11 +87,9 @@ def cast_from_epics_type(datatype: DataType[T], value: object) -> T:
     match datatype:
         case Enum():
             if len(datatype.members) <= MBB_MAX_CHOICES:
-                # epics type is mbb
                 return datatype.validate(datatype.members[value])
             else:  # enum backed by string record
-                # epics type is string
-                return datatype.validate(datatype.enum_cls(value))
+                return datatype.validate(datatype.enum_cls[value])
         case datatype if issubclass(type(datatype), EPICS_ALLOWED_DATATYPES):
             return datatype.validate(value)  # type: ignore
         case _:
@@ -105,7 +103,7 @@ def cast_to_epics_type(datatype: DataType[T], value: T) -> object:
             if len(datatype.members) <= MBB_MAX_CHOICES:
                 return datatype.index_of(datatype.validate(value))
             else:  # enum backed by string record
-                return datatype.validate(value).value
+                return datatype.validate(value).name
         case datatype if issubclass(type(datatype), EPICS_ALLOWED_DATATYPES):
             return datatype.validate(value)
         case _:
--- a/src/fastcs/transport/epics/ca/util.py
+++ b/src/fastcs/transport/epics/ca/util.py
@@ -86,7 +86,12 @@ def cast_from_epics_type(datatype: DataType[T], value: object) -> T:
     """Casts from an EPICS datatype to a FastCS datatype."""
     match datatype:
         case Enum():
-            return datatype.validate(datatype.members[value])
+            if len(datatype.members) <= MBB_MAX_CHOICES:
+                # epics type is mbb
+                return datatype.validate(datatype.members[value])
+            else:  # enum backed by string record
+                # epics type is string
+                return datatype.validate(datatype.enum_cls(value))
         case datatype if issubclass(type(datatype), EPICS_ALLOWED_DATATYPES):
             return datatype.validate(value)  # type: ignore
         case _:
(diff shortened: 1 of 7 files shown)


GDYendell (approved):
Thanks for spotting this. I guess this was broken in #102

GDYendell:
Looks good. Happy to merge when the coverage is bumped up.

Could you add three parameterized tests for cast_from_epics_type, cast_to_epics_type, builder_callable_from_attribute to cover all the branches?

jsouter:
Aha, so this of course only works when we are using a string enum, which is what I had in my fastcs-eiger branch. I guess we need to decide if we handle several different types of enum based on the underlying type or if we restrict ourselves.
Currently FastCS expects simple Enums with ints as the underlying type. We could require, or support alongside, string type enums, this would allow us to have labels with spaces and other characters as per https://github.com/DiamondLightSource/FastCS/issues/149 - we would use the Enum member's value when constructing guis. For int enums, the labels are obtained from the Enum member's name.

jsouter:
Proposal: something like
```
class Enum(Generic[T_Enum], DataType[T_Enum]):
    enum_cls: type[T_Enum]

    def __post_init__(self):
        if not issubclass(self.enum_cls, enum.Enum):
            raise ValueError("Enum class has to take an Enum.")

    def index_of(self, value: T_Enum) -> int:
        return self.members.index(value)

    def label_of(self, value: T_Enum):
        if self._is_string:
            return value.value
        else:
            return value.name

    @cached_property
    def members(self) -> list[T_Enum]:
        return list(self.enum_cls)

    @cached_property
    def _is_string(self):
        return all(isinstance(member.value, str) for member in self.members)

    @cached_property
    def labels(self) -> list[str]:
        return [self.label_of(member) for member in self.members]

    @property
    def dtype(self) -> type[T_Enum]:
        return self.enum_cls

    @property
    def initial_value(self) -> T_Enum:
        return self.members[0]
```
Then when we populate the metadata/give choices in the dropdowns we use the labels, which are the string values if all members of the enum have str values, otherwise the names of the members.

evvaaaa:
### Leaving 

> Seeing from example_p4p_ioc.py that Eva intended the enums to be able to have arbitrary values at least for p4p, so maybe this restriction needs more thought.

For both p4p and softioc, even if the `.value` in the `enum_cls` is an integer, we still use either the string `.name` or the index of the `.name` in the `list(enum_cls)`.

My suggestion would be that the string values you would put in the case of either an `mbbOut`, or `stringOut` (for options `n>16`), are the names of the elements.

I would suggest in the case of `n>16`, we do a `logging.warning` and state that only the string names can be used in the record, not integer indices.

jsouter:
> For both p4p and softioc, even if the `.value` in the `enum_cls` is an string/integer, we still use the string `.name` / the index of the `.name` in the `list(enum_cls)` when writing from the client.

Right okay, so the `.value` only has meaning downstream. I still think we need some mechanism to specify more flexible string labels for the dropdowns, whether or not it affects the underlying epics value.

evvaaaa:
> Maybe we could add a helper for constructing these?

I think the clearest thing would be to just use a normal python enum and access it with standard `[]`:

``` python
Foo = Enum("Foo", {"One option": 1, "Second option": "2"})

print(Foo["One option"], Foo["One option"].value)
print(Foo["Second option"], Foo["Second option"].value)
```

```
Foo.One option 1
Foo.Second option 2
```

Okay, ich bin aus 🫡

evvaaaa:
Also, as part of this you might want to add an `on_update`, `before=True` which rejects the put on the `longStringOut` if it isn't a string name in the enum.

Without this, you would put `"Some value not in the record enum.Enum"` to the record, it would be accepted by softioc, rejected by the `Attribute`.

jsouter:
> Also, as part of this you might want to add an `on_update`, `before=True` which rejects the put on the `longStringOut` if it isn't a string name in the enum.
> 
> Without this, you would put `"Some value not in the record enum.Enum"` to the record, it would be accepted by softioc, rejected by the `Attribute`.

Yep, that's a good idea.

src/fastcs/transport/epics/ca/util.py
```diff
@@ -78,6 +80,12 @@ def record_metadata_from_datatype(datatype: DataType[T]) -> dict[str, str]:
                     )
                 )
                 arguments.update(state_keys)
+            elif out_record:  # no validators for in type records
+
+                def _verify_in_datatype(_, value):
+                    return value in [member.name for member in datatype.members]
+
+                arguments["validate"] = _verify_in_datatype
```
GDYendell, line 88: This is OK for now, but I am slightly concerned that this mechanism of passing callables and kwargs around to make record instantiation generic and DRY is becoming unreadable.

GDYendell (approved):

PR #158, opened 2025-05-23
Allow creating CA IOCs via yaml config

### LINKED ISSUE #145: It is not possible to create a CA IOC from yaml now that PVA has been added because the schema is the same

GDYendell:
This enables creating either ca or pva IOCs via yaml config by giving them uniquely named fields.

Fixes #145


LuisFSegalla (approved):
Seems good to me!
Managed to run all the tests and couldn't find anything wrong or that needed change

PR #163, opened 2025-06-11
Update installation check in tutorial

GDYendell:


gilesknap (approved):
Yep that works for me.

PR #164, opened 2025-06-23
Add GUI code to tutorial snippets

JamesDoingStuff:
- Adds GUI creation code to the tutorial snippets so that the `demo.bob` file is produced as expected. This is added as a new snippet (`snippet05.py`).
- Also makes small clarifications about the simulated temperature controller and why the last line gets commented out.

Note: After adding in the new snippet, I shifted the other snippet numbers to match. Git didn't like this much. So, while it may look like I made significant changes to all the snippets, this is actually just a consequence of the renaming. The only actual content changes are the gui creation lines towards the end of each one.

### DIFF
--- a/docs/tutorials/static-drivers.md
+++ b/docs/tutorials/static-drivers.md
@@ -379,7 +379,7 @@ controller.
 :class: dropdown, hint
 
 :::{literalinclude} /snippets/static13.py
-:emphasize-lines: 3,17,47,102-108
+:emphasize-lines: 3,17,102-108
 :::
 
 ::::
--- a/docs/snippets/static09.py
+++ b/docs/snippets/static09.py
@@ -40,7 +40,7 @@ class TemperatureControllerHandler(AttrHandlerRW):
 
     async def put(self, attr: AttrW, value: Any):
         await self.controller.connection.send_command(
-            f"{self.command_name}={value}\r\n"
+            f"{self.command_name}={attr.dtype(value)}\r\n"
         )
 
 
--- a/docs/snippets/static10.py
+++ b/docs/snippets/static10.py
@@ -40,7 +40,7 @@ class TemperatureControllerHandler(AttrHandlerRW):
 
     async def put(self, attr: AttrW, value: Any):
         await self.controller.connection.send_command(
-            f"{self.command_name}{self.controller.suffix}={value}\r\n"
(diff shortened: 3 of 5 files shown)


docs/tutorials/static-drivers.md
```diff
@@ -100,6 +97,11 @@ passing it in.
 
 ::::
 
+:::{note}
+In the above snippet and all hereafter, the final line is commented out. This is done to 
+avoid blocking our unit tests - in your own code, it should remain uncommented.
+:::
```
GDYendell, line 103: This is very helpful. Maybe it should be a warning so it is a bit more visible?

docs/tutorials/static-drivers.md
```diff
@@ -215,11 +231,11 @@ The set commands do not return a response, so use the `send_command` method inst
...
 :::
 
-::::{admonition} Code 8
+::::{admonition} Code 9
 :class: dropdown, hint
 
-:::{literalinclude} /snippets/static08.py
-:emphasize-lines: 4,6,16,40-43,49
+:::{literalinclude} /snippets/static09.py
+:emphasize-lines: 5,7,17,41-44, 48-50
```
GDYendell, line 238: 48 and 49 are the same as the last snippet.
```suggestion
:emphasize-lines: 5,7,17,41-44,50
```

docs/snippets/static13.py
```diff
@@ -0,0 +1,122 @@
...
+        response = await self.controller.connection.send_query(
+            f"{self.command_name}{self.controller.suffix}?\r\n"
+        )
+        value = response.strip("\r\n")
+
+        await attr.set(attr.dtype(value))
+
+    async def put(self, attr: AttrW, value: Any) -> None:
+        await self.controller.connection.send_command(
+            f"{self.command_name}{self.controller.suffix}={attr.dtype(value)}\r\n"
```
GDYendell, line 47: The `attr.dtype()` should be used the first time this put method appears really

GDYendell (changes requested):
Thanks @JamesDoingStuff! Just a few minor changes.

docs/tutorials/static-drivers.md
JamesDoingStuff: Good idea, made the change

PR #168, opened 2025-07-07
Add cast to Bool for EPICS bool records

GDYendell:


jsouter (approved):
Looks good, I'd seen this error with fastcs-odin but didn't realise it was a coming from the CA backend, assumed it was some weirdness on the odin side.

PR #170, opened 2025-07-18
Add RBV signals to PVA transport

### LINKED ISSUE #153: Create _RBV records for PVA transport
Following discussion in #152
We should create separate read and write PVs for signals in PVA as we do in CA. 

## Acceptance Criteria
- Create _RBV records for RW signals in PVA
- Generate phoebus screens that are fully compatible with PVA
- Write an ADR to explain this change in design

shihab-dls:
Fixes #153 

This PR splits out `make_shared_pv` for read and write signals, creating two shared PV instances for `AttrRW`, where one is a `_RBV`.

### DIFF
--- a/src/fastcs/transport/epics/pva/_pv_handlers.py
+++ b/src/fastcs/transport/epics/pva/_pv_handlers.py
@@ -13,6 +13,7 @@ from fastcs.datatypes import Table
 from .types import (
     MAJOR_ALARM_SEVERITY,
     RECORD_ALARM_STATUS,
+    T,
     cast_from_p4p_value,
     cast_to_p4p_value,
     make_p4p_type,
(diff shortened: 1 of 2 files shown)


tests/transport/epics/pva/test_p4p.py
```diff
@@ -464,10 +464,12 @@ async def _wait_and_put_pvs():
...
     waveform_values, table_values, enum_values = [], [], []
-    waveform_monitor = ctxt.monitor(f"{pv_prefix}:SomeWaveform", waveform_values.append)
-    table_monitor = ctxt.monitor(f"{pv_prefix}:SomeTable", table_values.append)
+    waveform_monitor = ctxt.monitor(
+        f"{pv_prefix}:SomeWaveform_RBV", waveform_values.append
+    )
+    table_monitor = ctxt.monitor(f"{pv_prefix}:SomeTable_RBV", table_values.append)
     enum_monitor = ctxt.monitor(
-        f"{pv_prefix}:SomeEnum",
+        f"{pv_prefix}:SomeEnum_RBV",
```
GDYendell, line 472: Do the monitor queues get populated the same if monitoring the demand vs the rbv?

shihab-dls: Seems that the queues are only populated with the client put if we **don't** use the `_RBV` monitor; otherwise both the client and server puts are populated.

shihab-dls: Just pushed some comments to clarify this; let me know if they are sufficient! (re-requesting review)

PR #171, opened 2025-07-23
Ensure group is passed when binding commands

JamesOHeaDLS:


GDYendell:
Could you update [this test](https://github.com/DiamondLightSource/FastCS/blob/4bea912278d67a3c0a21079394f8396ec9cccbef/tests/test_cs_methods.py#L40) to pass `group` to `UnboundCommand` and assert that the command has the same group set?

JamesOHeaDLS:
Hi @GDYendell 
I think this does what you are after.  At least, if I remove the fix to cs_methods.py it fails
Thanks,
James

GDYendell (approved):

PR #174, opened 2025-08-04
Allow enum datatypes to be set to their enum values

### LINKED ISSUE #173: Allow enum attributes to be set to their values
The `Enum` `DataType` currently doesn't have its own validate function, so it uses the default default parent validation which errors `if not isinstance(value, self.dtype)`. This means if I have an enum like
```
Class MyEnum(Enum):
    MY_VAL = "Val"
```
then the only values that are accepted when doing a `put` into an attribute of this type are instances of this enum. For instance, I can do `await attr.set(MyEnum.MY_VAL)`, but not `await attr.set("Val"). 

It would be useful to extend this validation so that we can set these attributes to their enum values.

## Acceptance Criteria
- Add a validation function to the enum datatype to allow enum members to be used in the `set`
- No breaking changes occur - you can still use `set` with enum instances as before

olliesilvester:
Fixes #173

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=64", "setuptools_scm[toml]>=8"]
+requires = ["setuptools>=70.1", "setuptools_scm[toml]>=8"]
 build-backend = "setuptools.build_meta"
 
 [project]
@@ -20,7 +20,7 @@ dependencies = [
     "pytango",
     "softioc>=4.5.0",
     "strawberry-graphql",
-    "p4p"
+    "p4p",
 ]
 dynamic = ["version"]
 license.file = "LICENSE"
@@ -53,9 +53,7 @@ dev = [
     "httpx",
     "tickit~=0.4.3",
 ]
-demo = [
-    "tickit~=0.4.3",
-]
+demo = ["tickit~=0.4.3"]
 
 [project.scripts]
 fastcs-demo = "fastcs.demo.__main__:main"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=64", "setuptools_scm[toml]>=8"]
+requires = ["setuptools>=70.1", "setuptools_scm[toml]>=8"]
```
olliesilvester, line 2: This version was required for pytest to work in a devcontainer

src/fastcs/datatypes.py
```diff
@@ -139,6 +139,17 @@ def __post_init__(self):
     def index_of(self, value: T_Enum) -> int:
         return self.members.index(value)
 
+    def validate(self, value: T) -> T:
+        enum_vals = [key.value for key in self.dtype]
+
+        if value not in enum_vals and not issubclass(type(value), self.dtype):
```
jsouter, line 145: can we avoid creating enum_vals with
`if value not in self.dtype`?

olliesilvester: The [__contains__ method for an enum](https://docs.python.org/3/library/enum.html#enum.EnumType.__contains__) was only added in python 3.12 so we can't do this until we drop 3.11 :(

jsouter: :disappointed:

GDYendell: [Soon!](https://scientific-python.org/specs/spec-0000/#support-window)

PR #184, opened 2025-08-09
Add helm chart for deploying fastcs IOC instances

### LINKED ISSUE #177: Create a helm chart for fast CS IOCs
- this should be published as an OCI chart by CI of this repo
- allows python projects that use fastCS to be deployed into a cluster
- question - could this instead be a helm chart for python-copier template in general

gilesknap:
- Also see the fastcs-example project here https://github.com/DiamondLightSource/fastcs-example
- And here is an example deployment of the `fastcs-example` using the published helm chart `0.9.2-alpha.3 
https://github.com/gilesknap/t01-services/tree/add-fastcs-example/services/bl01t-ea-fastcs-01

Resolves #177


gilesknap:
@GDYendell , I think I prefer to generalize this to have structure like you did in odin and I did in ec-helm-charts:

- Charts/fastcs-instance

Even though we only have one chart here.
That way we can make general purpose actions that can be dropped in to any project.

Waddyathink?

.github/workflows/_helm.yml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Push tagged helm chart to registry
+        if: ${{ github.ref_type == 'tag' }}
+        run: |
+          set -x
+
+          echo "${{ secrets.GITHUB_TOKEN }}" | helm registry login ghcr.io/${{ github.repository_owner }} --username ${{ github.repository_owner }} --password-stdin
+          REGISTRY=oci://ghcr.io/diamondlightsource/charts
+          for i in charts/*.tgz; do
+            helm push "${i}" ${REGISTRY}
+          done
```
GDYendell, line 71: This should go in the release job

Charts/fastcs-instance/templates/service.yaml
```diff
@@ -0,0 +1,52 @@
...
+    {{- include "fastcs.labels" . | nindent 4 }}
+    app: {{ .Release.Name }}
+    location: {{ .Values.global.location }}
+    ioc_group: {{ .Values.global.ioc_group }}
+    is_ioc: "true"
+spec:
+  selector:
+    app: {{ .Release.Name }}
+  type: {{ .Values.service.type }}
+  # allocate a fixed clusterIP for this service based on the service name
```
GDYendell, line 15: Why?

Charts/fastcs-instance-ioc/Chart.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v2
+name: fastcs-instance-ioc
```
GDYendell, line 2: This name is a bit confusing. Could we just have `fastcs` and `fastcs-ioc`?

schemas/fastcs-instance.schema.json
```diff
@@ -0,0 +1 @@
+../Charts/fastcs-instance/values.schema.json
```
GDYendell, line 1: Do we need to check in the schemas or can we just publish them in CI?

GDYendell (commented):
I am a bit confused by the two helm charts. Is the intention that there is a base chart that doesn't assume what transport you are using and then an EPICS specific one? Because currently the base chart has EPICS things in it. If not, what are the two charts for?

.github/workflows/_helm.yml
gilesknap: I like it not in the release job because:
  - it keeps the helm related things in a single workflow
  - there is very little connection between what the project releases and its helm chart as its intended for use in services repos that want to include projects that have fastcs as a dependency
 
this does beg the question, should the helm chart even be in this repo?

GDYendell: It could be in its own release job, but it should depend on the other test jobs working. Really it should be the very last thing because it depends on all other releases jobs working too - no good having the helm chart if the container didn't publish for some reason.

Maybe the charts could be in their own repo... It does risk them going out of sync though.

.github/workflows/_helm.yml
```diff
@@ -0,0 +1,71 @@
...
+  workflow_call:
+
+env:
+  HELM_VERSION_TO_INSTALL: 3.14.3
+
+jobs:
+  package-helm-charts:
+    name: Package and Push Helm Chart
+    runs-on: ubuntu-latest
+    environment: release
```
GDYendell, line 13: Could we remove this so it doesn't spam PRs with deployment messages? Maybe we could keep it if the publish was in its own job, but I am not sure it is adding much.

Charts/fastcs-instance/templates/service.yaml
```diff
@@ -0,0 +1,52 @@
...
+      targetPort: {{ .Values.pva_server_port | default 5075 }}
+      protocol: UDP
+    - name: pva-broadcast-tcp
+      port: {{ add1 (.Values.pva_server_port | default 5075) }}
+      targetPort: {{ add1 (.Values.pva_server_port | default 5075) }}
+      protocol: TCP
+    - name: pva-broadcast-udp
+      port: {{ add1 (.Values.pva_server_port | default 5075) }}
+      targetPort: {{ add1 (.Values.pva_server_port | default 5075) }}
+      protocol: UDP
```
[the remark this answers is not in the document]
GDYendell: You can make a library helm chart to define snippets that can then be included in other charts. There are some examples [here](https://github.com/bitnami/charts/tree/main/bitnami/common/templates).

gilesknap:
> This has also made me wonder how easy it is going to be to deploy multiple transports, because each is going to need their own ports exposed etc.

My take is that we just publish ports for all transports in the one helm chart - so far I have PVA and CA in there.

Or we could pass config to the service to expose ports as needed, that just means we specify the list of transports in 2 places.

BTW. At DLS we are still using hostNetwork so this service is not used except in demos and training.

PR #185, opened 2025-08-11
Do not propagate DRVL and DRVH to in type records in CA transport

### LINKED ISSUE #172: DRVL and DRVH is incorrectly set on in records
Currently if min/max is set on an attribute, in the EPICS transport it sets DRVL and DRVH on both in and out records. This should only be done for out records.

## Steps To Reproduce
1. Run and IOC with a Controller with and Int(..., max=10)

## Acceptance Criteria
- No error
- out records have DRVL and DRVH passed through
- in records do not

cc @JamesOHeaDLS

jsouter:
Closes #172

### DIFF
--- a/tests/transport/epics/ca/test_softioc.py
+++ b/tests/transport/epics/ca/test_softioc.py
@@ -541,7 +541,7 @@ def test_update_datatype(mocker: MockerFixture):
         attr_r.update_datatype(String())  # type: ignore
 
     attr_w = AttrW(Int())
-    record_w = _make_record(pv_name, attr_w, on_update=mocker.ANY)
+    record_w = _make_record(pv_name, attr_w, on_update=mocker.ANY, out_record=True)
 
     builder.longIn.assert_called_once_with(
         pv_name,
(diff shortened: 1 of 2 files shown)


tests/transport/epics/ca/test_softioc.py
```diff
@@ -547,9 +549,9 @@ def test_update_datatype(mocker: MockerFixture):
         **record_metadata_from_datatype(attr_w.datatype),
     )
     record_w.set_field.assert_not_called()
-    attr_w.update_datatype(Int(units="m", min=-3))
+    attr_w.update_datatype(Int(units="m", min_alarm=-3))
     record_w.set_field.assert_any_call("EGU", "m")
-    record_w.set_field.assert_any_call("DRVL", -3)
+    record_w.set_field.assert_any_call("LOPR", -3)
```
GDYendell, line 554: Could we keep this as it was to show that min and DRVL are valid for AttrW, with the AttrR check above validating min_alarm -> LOPR?

PR #186, opened 2025-08-11
Use SimpleHandler on AttrRW only

### LINKED ISSUE #126: Only use SimpleHandler for AttrRW
Currently `SimpleHandler` is used as a default handler if none is given for `AttrW`, but this doesn't make sense.`SimpleHandler` is intended for creating attributes that can be set and just store the value for internal use. `AttrW` is specifically used for sending something to a device and not storing any state, so the value cannot be read.

The default is set [here](https://github.com/DiamondLightSource/FastCS/blob/98807bc290d76e6773517d699d5d030a53675e13/src/fastcs/attributes.py#L182), and this should be moved to `AttrRW`. The handler can also be updated to explicitly only take `AttrRW` to remove the `isinstance` check.

It would also be good to add a test that makes a controller with an `AttrRW` with no handler given and then sets the value and reads it back.

## Acceptance Criteria
- `SimplerHandler` is only used by default for `AttrRW`

shihab-dls:
fixes #126 

This removes use of `SimpleHandler` from `AttrW` and defaults it in `AttrRW`. Testing `AttrW` with this handler has been removed, and testing with `AttrRW` amended to check we can read the value we set.

### DIFF
--- a/src/fastcs/attributes.py
+++ b/src/fastcs/attributes.py
@@ -49,11 +49,9 @@ class AttrHandlerRW(AttrHandlerR, AttrHandlerW):
 class SimpleAttrHandler(AttrHandlerRW):
     """Handler for internal parameters"""
 
-    async def put(self, attr: AttrW[T], value: T) -> None:
+    async def put(self, attr: AttrRW[T], value: T) -> None:  # type: ignore
+        await attr.set(value)
         await attr.update_display_without_process(value)
-
-        if isinstance(attr, AttrRW):
-            await attr.set(value)
 
     async def update(self, attr: AttrR) -> None:
         raise RuntimeError("SimpleHandler cannot update")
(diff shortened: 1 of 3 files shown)


src/fastcs/attributes.py
```diff
@@ -49,12 +49,10 @@ class AttrHandlerRW(AttrHandlerR, AttrHandlerW):
 class SimpleAttrHandler(AttrHandlerRW):
     """Handler for internal parameters"""
 
-    async def put(self, attr: AttrW[T], value: T) -> None:
+    async def put(self, attr: AttrRW[T], value: T) -> None:  # type: ignore
```
GDYendell, line 52: Ah, OK. We can't remove the `isinstance` because `put` needs to take an `AttrW`. Let's keep it
as it was.

GDYendell (changes requested):
Just one comment

GDYendell (approved):

PR #188, opened 2025-08-12
Add API to update attributes once when Backend.serve() called

### LINKED ISSUE #176: Add an API for a scan and a AttrR to only poll once with a ONCE constant
Some attributes need to be updated from the device, but the value will never change. Add a special ONCE constant and update scan and AttrR.update_period to take a float or ONCE. If it is ONCE, add the update callback to initial scan tasks instead of making a task to call it periodically.

## Acceptance Criteria
- Specific criteria that will be used to judge if the issue is fixed

jsouter:
Closes #176
draft, ONCE is just an empty class at the moment as proof of concept (has to be called like `ONCE()`)


docs/snippets/dynamic.py
```diff
@@ -59,7 +66,7 @@ def create_attributes(parameters: dict[str, Any]) -> dict[str, Attribute]:
 @dataclass
 class TemperatureControllerHandler(AttrHandlerRW):
     command_name: str
-    update_period: float | None = 0.2
+    update_period: UpdatePeriod = 0.2
```
GDYendell, line 69: Does the type hint have to be `UpdatePeriod` to make the type checker happy? It might be nicer to keep this union definition private.

PR #189, opened 2025-08-12
Add generic interactive shell

### LINKED ISSUE #140: Create a generic interactive shell and allow transports to add to the context
This will allow us to do `dbl()` again, amongst other things.

shihab-dls:
fixes #140 

This PR adds an interactive shell, based on this [implementation](https://github.com/GDYendell/fastcs2/blob/main/src/fastcs2/control_system.py#L44), and adds transport contexts. The `wip` commits are attempts to add `dbl` `dbpf` and `dbgf` functionality to `PVA` as an example, but maybe only the first commit is required for this PR, and further PRs can add transport specific context.

### DIFF
--- a/src/fastcs/launch.py
+++ b/src/fastcs/launch.py
@@ -1,6 +1,7 @@
 import asyncio
 import inspect
 import json
+import os
 import signal
 from collections.abc import Coroutine
 from functools import partial
@@ -104,12 +105,21 @@ class FastCS:
 
     async def serve(self) -> None:
         coros = [self._backend.serve()]
-        if self._transports:
-            context = {"controller": self._backend}
-            for transport in self._transports:
-                context.update(transport.context())
-                coros.append(transport.serve())
+        context = {
+            "controller": self._controller,
+            "controller_api": self._backend.controller_api,
+            "transports": [
+                transport.__class__.__name__ for transport in self._transports
+            ],
+        }
+
+        for transport in self._transports:
+            coros.append(transport.serve())
+            context.update(transport.context())
+
+        if not os.getenv("NOT_INTERACTIVE"):
             coros.append(self._interactive_shell(context))
+
         try:
             await asyncio.gather(*coros)
         except asyncio.CancelledError:
--- a/src/fastcs/launch.py
+++ b/src/fastcs/launch.py
@@ -104,11 +104,12 @@ class FastCS:
 
     async def serve(self) -> None:
         coros = [self._backend.serve()]
-        context = {"controller": self._backend}
-        for transport in self._transports:
-            context.update(transport.context())
-            coros.append(transport.serve())
-        coros.append(self._interactive_shell(context))
+        if self._transports:
+            context = {"controller": self._backend}
+            for transport in self._transports:
+                context.update(transport.context())
+                coros.append(transport.serve())
+            coros.append(self._interactive_shell(context))
         try:
             await asyncio.gather(*coros)
         except asyncio.CancelledError:
--- a/src/fastcs/launch.py
+++ b/src/fastcs/launch.py
@@ -1,7 +1,6 @@
 import asyncio
 import inspect
 import json
-import os
 import signal
 from collections.abc import Coroutine
 from functools import partial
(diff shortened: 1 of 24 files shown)


src/fastcs/launch.py
```diff
@@ -100,12 +104,43 @@ def run(self):
 
     async def serve(self) -> None:
         coros = [self._backend.serve()]
-        coros.extend([transport.serve() for transport in self._transports])
+        if self._transports:
+            context = {"controller": self._backend}
+            for transport in self._transports:
+                context.update(transport.context())
```
GDYendell, line 110: We should have a checked update that errors if a key already exists. Hopefully it will never come up, but if it does we need to know.

src/fastcs/transport/adapter.py
```diff
@@ -22,3 +22,7 @@ def create_docs(self) -> None:
     @abstractmethod
     def create_gui(self) -> None:
         pass
+
+    @abstractmethod
+    def context(self) -> dict[str, Any]:
+        return {}
```
GDYendell, line 28: Make this not abstract so transports don't have to implement it

src/fastcs/transport/adapter.py
```diff
@@ -22,3 +22,7 @@ def create_docs(self) -> None:
     @abstractmethod
     def create_gui(self) -> None:
         pass
+
+    @abstractmethod
+    def context(self) -> dict[str, Any]:
```
GDYendell, line 27: This should be a property because it is not a verb

GDYendell (changes requested):
Minor comments

src/fastcs/launch.py
```diff
@@ -100,12 +104,43 @@ def run(self):
 
     async def serve(self) -> None:
         coros = [self._backend.serve()]
-        coros.extend([transport.serve() for transport in self._transports])
+        if self._transports:
+            context = {"controller": self._backend}
```
GDYendell, line 108:
```suggestion
            context = {"controller": self._controller}
```

src/fastcs/launch.py
```diff
@@ -100,12 +105,52 @@ def run(self):
...
+                transport.__class__.__name__ for transport in self._transports
+            ],
+        }
+
+        for transport in self._transports:
+            coros.append(transport.serve())
+            context.update(transport.context)
+
+        if not os.getenv("NOT_INTERACTIVE"):
+            coros.append(self._interactive_shell(context))
```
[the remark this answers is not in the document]
shihab-dls: Hmm, we could make a `TestShell` fixture that hangs, and inject this into `FastCS.__init__`, storing the Iterminal in `self._ip_shell`? Something like:
```python
@pytest.fixture
def test_shell():
    class TestShell:
        def __init__(self):
            self._stop = threading.Event()

        def mainloop(self):
            self._stop.wait()

        def exit(self):
            self._stop.set()

    return TestShell()
```
We could then patch this like:
```python
@pytest.fixture(autouse=True)
def patch_fastcs_init_with_test_shell(test_shell):
    def wrapped_init(self, *args, **kwargs):
        fastcs.launch.FastCS.__init__(self, *args, **kwargs)
        self._ip_shell = test_shell

    with patch.object(fastcs.launch.FastCS, "__init__", wrapped_init):
        yield
```
I'm not sure if this is what you were going for though? I guess we would then need to find specific fastcs events in any test that executes `fastcs.run()` (which currently is the doc snippets) we want to trigger `test_shell.exit()`, or we could add another fixture that does `threading.Timer(delay, test_shell.exit)` to exit a bit after yielding to the test, although an event driven approach is probably better.

shihab-dls:
I've commented out `fastcs.run()` lines that were in some doc snippets for now, and removed the env var. Re-requesting a review with the generic terminal and the amended test. I've also added a test for the duplicate context error for coverage; let me know if you would've tested it another way.

Codecov is still failing, but for reasons I currently think are "okay"...

GDYendell:
I think I am OK with leaving that run method with missing lines unless you can come up with a neat way to test that

shihab-dls:
> I think I am OK with leaving that run method with missing lines unless you can come up with a neat way to test that

Hit my head against it for a little while, and can't figure out a "neat" way to do it; I can open up a small issue for me to look at after, but for now we should probably merge this in (if you are happy with my other test)?

PR #192, opened 2025-08-15
Implement SubControllerVector

### LINKED ISSUE #122: Allow for explicit pairing of SubControllers/Attributes
In https://github.com/DiamondLightSource/FastCS/pull/119, we pair attributes/controllers together based on the name of the attribute. If the name of two attributes/subcontrollers `foo_1`/`foo_2` or `Foo1`/`Foo2` is the same then they're grouped together in the same pvi group under `foo`.

# Problem

This is assuming too much on the FastCS side. We should instead allow controllers to have a `AttributeVector`/`ControllerVector` called `foo` which can be indexed by `int` (not necessarily contiguous). On the transport, these elements are grouped together under `foo`.

# Required Changes

1. Add a new `AttributeVector` type.
2. Add a new `ControllerVector` which can be passed into `register_subcontroller`.

shihab-dls:
fixes #122 

This PR introduces a `SubControllerVector`, which is a mapping of `int` to `SubController`, used in pvi grouping.

### DIFF
--- a/src/fastcs/controller.py
+++ b/src/fastcs/controller.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+from collections.abc import Iterator, Mapping, MutableMapping
 from copy import deepcopy
 from typing import get_type_hints
 
@@ -27,6 +28,7 @@ class BaseController:
             self.attributes = {}
         self._path: list[str] = path or []
         self.__sub_controller_tree: dict[str, SubController] = {}
+        self.__sub_controller_vector: str | None = None
 
         self._bind_attrs()
 
@@ -98,7 +100,15 @@ class BaseController:
             elif isinstance(attr, UnboundPut | UnboundScan | UnboundCommand):
                 setattr(self, attr_name, attr.bind(self))
 
-    def register_sub_controller(self, name: str, sub_controller: SubController):
+    def register_sub_controller(
+        self, name: str, sub_controller: SubController | SubControllerVector
+    ):
+        if isinstance(sub_controller, SubControllerVector):
+            for index, child in sub_controller.children():
+                child.__sub_controller_vector = name  # noqa: SLF001
+                self.register_sub_controller(f"{name}{index}", child)
+            return
+
         if name in self.__sub_controller_tree.keys():
             raise ValueError(
                 f"Controller {self} already has a SubController registered as {name}"
@@ -118,6 +128,9 @@ class BaseController:
 
     def get_sub_controllers(self) -> dict[str, SubController]:
         return self.__sub_controller_tree
+
+    def get_sub_controller_vector(self) -> str | None:
+        return self.__sub_controller_vector
 
 
 class Controller(BaseController):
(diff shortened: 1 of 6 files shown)


shihab-dls:
Unsure about codecov currently, so will request a review to make sure the pvi structure and PV formatting is what we expect

src/fastcs/controller.py
```diff
@@ -199,8 +201,74 @@ def __init__(
...
+        description: str | None = None,
+        ios: Sequence[AttributeIO[T, AttributeIORefT]] | None = None,
+    ) -> None:
+        self._children: dict[int, Controller] = {}
+        self.update(children)
+        super().__init__(description=description, ios=ios)
+        for index, child in children.items():
+            super().add_sub_controller(str(index), child)
+
+    def add_sub_controller(self, *args, **kwargs):
```
GDYendell: I think pyright won't like not having the same function signature here?

src/fastcs/transport/epics/pva/pvi_tree.py
```diff
@@ -40,16 +42,19 @@ class PviDevice(dict[str, "PviDevice"]):
     """For creating a pvi structure in pva."""
 
     pv_prefix: str
+    controller_api: ControllerAPI | None
     description: str | None
     device_signal_info: _PviSignalInfo | None
```
GDYendell: I think these are redundant?

src/fastcs/transport/epics/pva/pvi_tree.py
```diff
@@ -190,6 +203,9 @@ def add_sub_device(
             self._pvi_tree_root.get_recursively(
                 *device_pv.split(":")[1:]  # To remove the prefix
             ).description = description
+            self._pvi_tree_root.get_recursively(
+                *device_pv.split(":")[1:]  # To remove the prefix
+            ).controller_api = controller_api
```
GDYendell: Maybe not for this PR, but I think we should remove `get_recursively` and instead have `add_sub_device` return the next `PviDevice`. Then we can explicitly walk down the controller api, adding signals for attributes and creating more sub devices for sub controllers. I am also not sure `PviTree` and `PviDevice` should be separate classes.

What do you think?

src/fastcs/controller.py
```diff
@@ -199,8 +201,74 @@ def __init__(
...
+
+
+class ControllerVector(MutableMapping[int, Controller], BaseController):
+    """A collection of SubControllers, with an arbitrary integer index.
+    An instance of this class can be registered with a parent ``Controller`` to include
+    it's children as part of a larger controller. Each child of the vector will keep
+    a string name of the vector.
+    """
+
+    root_attribute: Attribute | None = None
```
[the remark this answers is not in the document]
shihab-dls: `ControllerVector` no longer inherits from `Controller` though; It inherits from `BaseController`

src/fastcs/transport/epics/ca/util.py
```diff
@@ -91,6 +91,9 @@ def _verify_in_datatype(_, value):
                     return value in datatype.names
 
                 arguments["validate"] = _verify_in_datatype
+        case Bool():
+            arguments["ZNAM"] = "False"
+            arguments["ONAM"] = "True"
```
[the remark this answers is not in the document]
shihab-dls: I'm actually not sure how this got here; looks like you committed these in [this](https://github.com/DiamondLightSource/FastCS/pull/192/commits/7591e8fa87700b98c18b882febab955753d92148) ref? Not sure if this relates to another ticket?

src/fastcs/transport/epics/pva/pvi_tree.py
shihab-dls: I've included these suggestions in this PR, alongside your review comment, so I'm re-requesting a review now (sorry for the 27 commits)

src/fastcs/transport/epics/ca/util.py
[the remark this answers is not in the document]
GDYendell: Ah right yes and I couldn't rebase it away because there is a merge from main after it. If we merge that change first and then squash merge this it should be fine

src/fastcs/controller_api.py
```diff
@@ -38,9 +38,12 @@ def walk_api(self) -> Iterator["ControllerAPI"]:
             yield from api.walk_api()
 
     def __repr__(self):
-        return f"""\
-ControllerAPI(path={self.path}, sub_apis=[{", ".join(self.sub_apis.keys())}])\
-"""
+        return (
+            f"ControllerAPI("
+            f"path={self.path}, "
+            f"sub_apis=[{', '.join(sub_api for sub_api in self.sub_apis.keys())}]"
```
GDYendell: What does the `sub_api for sub_api in ` achieve here?

src/fastcs/transport/epics/ca/util.py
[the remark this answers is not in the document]
GDYendell: Made #261 to resolve this

PR #200, opened 2025-08-28
Add Helm Chart with debug capabilities and values schema generation using https://github.com/losisin/helm-values-schema-json

gilesknap:


Charts/README_CHARTS.md
```diff
@@ -0,0 +1,71 @@
...
+  - The contents of /workspaces and /venv are copied into the PVC.
+  - The venv from the debug image is an editable install of the project source code in /workspaces.
+  - The PVC folders are mounted over the corresponding folders in the container.
+- `autostart`:
+  - When false, the container starts with PID 1 as sleep infinity.
+  - When true, the container starts with its normal entrypoint.
+
+In combination these flags can be used to debug or develop in-cluster.
+
+An initial demonstration script to use these features is provided in [debug.py](https://github.com/epics-containers/p47-services/blob/add-fastcs/debug.py) in the `p47-services` repo.
```
GDYendell, line 59: If this is fastcs specific should we just put in in here as the reference version for now? Could it be made generic for all services by moving autostart and editable to globals?

Charts/README_CHARTS.md
```diff
@@ -0,0 +1,71 @@
...
+
+This script will:
+
+- inspect the values of `editable` and `autostart` in the `values.yaml` file of the specified IOC (TODO: at present it uses the p47-services source code to do so but this should be determined from the cluster in future).
```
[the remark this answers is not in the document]
GDYendell: And remove TODO

Charts/fastcs/values.yaml
```diff
@@ -0,0 +1,145 @@
...
+#     image:
+#       type: string
+#       format: image
+#       description: Container image URI
+#     command:
+#       type: array
+#       description: Command to run in the container (as array of arguments)
+#   required: [name, image, command]
+#   additionalProperties: false
+# @schema
```
GDYendell, line 108: I guess this is copied because using a ref doesn't work? Currently this doesn't make it into the schema. Could we either remove it entirely or use this before merge?

Charts/fastcs/Chart.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v2
+name: fastcs
+description: A Helm chart for Kubernetes
```
GDYendell, line 3:
```suggestion
description: A Helm chart for FastCS applications
```
?

Charts/fastcs/templates/statefulset.yaml
```diff
@@ -0,0 +1,180 @@
...
+        securityContext:
+          {{- toYaml . | nindent 12 }}
+        {{- end }}
+        {{- with .Values.initCommand }}
+        command:
+          {{- toYaml . | nindent 12 }}
+        {{- end }}
+        volumeMounts:
+          - name: {{ .Release.Name }}-develop
+            mountPath: /dest
```
GDYendell, line 74: Could we have a better name? Maybe `/debug`?

schemas/fastcs-service.schema.json
```diff
@@ -0,0 +1,12 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "type": "object",
+    "properties": {
+        "fastcs": {
+            "$ref": "fastcs-values.schema.json",
+            "type": "object",
+            "additionalProperties": false
+        }
+    },
```
GDYendell, line 10: Do we not need `globals` here?

Charts/fastcs/templates/statefulset.yaml
gilesknap: but you will not see /dest in the prod contianer, only in the init container that copies files to /dest and then exists.

gilesknap: I dont think debug describes it - we can debug non-editable containers. It is more accurate to call it develop or editable perhaps.

Charts/fastcs/values.yaml
gilesknap: OK this is now fixed nicely in 9be05b729d460f2ca60982d0f7b305841b3072bd as per https://github.com/losisin/helm-values-schema-json/issues/251#issuecomment-3237674115

Charts/fastcs/templates/statefulset.yaml
gilesknap: @GDYendell and @gilesknap discussed in person and agreed that /dest is the correct mount folder name in this context.

Charts/README_CHARTS.md
```diff
@@ -0,0 +1,71 @@
...
+This folder works in tandem with [_helm.yml](../.github/workflows/_helm.yml) github actions workflow which:
+
+- Validates the stricter form of SemVer tag that helm requires.
+- Finds all subfolders in the `Charts` directory and packages them as Helm charts.
+- Publishes the packaged charts to ghcr.io/${{github.repository_owner }}/charts/CHART_NAME
+- Uploads the contents of /schemas to artifacts (meaning they will be published to the release)
+
+This standalone Helm related workflow is independent of the rest of the workflows except that the _pypi workflow has _helm added to its `needs` in `ci.yml`, making sure we only publish to pypi with valid SemVer tags.
+
+TODO: for a project that publishes containers referred to by the helm chart, the `_containers.yml` workflow should be added to the `needs` of _helm in `ci.yml`.
```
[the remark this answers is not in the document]
gilesknap: done. https://github.com/DiamondLightSource/python-copier-template/issues/286

PR #209, opened 2025-09-10
validate hinted attributes on subcontrollers

jsouter:
Oversight from https://github.com/DiamondLightSource/FastCS/pull/187

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -8,7 +8,7 @@ from pydantic import ValidationError
 
 from fastcs.attributes import AttrR, AttrRW
 from fastcs.backend import Backend
-from fastcs.controller import Controller
+from fastcs.controller import Controller, SubController
 from fastcs.datatypes import Bool, Enum, Float, Int, String
 from fastcs.util import numpy_to_fastcs_datatype, snake_to_pascal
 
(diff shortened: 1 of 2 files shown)


tests/test_util.py
```diff
@@ -124,3 +124,26 @@ class ControllerWrongEnumClass(Controller):
...
+            subcontroller = ControllerWithWrongType()
+            self.register_sub_controller("MySubController", subcontroller)
+
+    with pytest.raises(RuntimeError) as excinfo:
+        Backend(TopController(), loop)
+
+    assert str(excinfo.value) == (
+        "Controller `ControllerWithWrongType` failed to introspect hinted attribute "
+        "`hinted_missing` during initialisation"
+    )
```
GDYendell, line 149: Might be neater

```suggestion
    with pytest.raises(RuntimeError, match="failed to introspect hinted attribute"):
        Backend(TopController(), loop)
```

GDYendell:
I am just doing some updates to fastcs-eiger and it has some logic to check `datatype` and `access_mode` of hinted attributes that get introspected. I think these checks should be done in `add_attribute` (where currently it fails immediately because the hinted attribute already exists) and `validate_hinted_attributes` should only check for hinted attributes that weren't introspected. Then in fastcs-eiger we can just call `add_attribute` and fastcs will do all the checks.

What do you think? Possibly this could be a separate PR.

PR #218, opened 2025-09-19
Replace Handler with AttributeIO / AttributeIORef

### LINKED ISSUE #211: Replace Handler with AttributeIO / AttributeIORef
The Handler classes currently create a circular dependency `Controller -> Attribute -> Handler -> Controller` and these all have to be defined in one file in drivers. They also cannot be statically linked and have to check at runtime if they have been passed the Controller they expect.

Apply the structure demonstrated [here](https://github.com/GDYendell/fastcs2) to improve this API.

- Create `AttributeIO` and `AttributeIORef` classes
- Update `Attribute*` to use these instead of `Handler`
- Update `Controller` binding logic to work with `AttributeIO`
- Remove `Handler*` classes

Links:
- [AttributeIO](https://github.com/GDYendell/fastcs2/blob/main/src/fastcs2/attribute_io.py)
- [AttributeIORef](https://github.com/GDYendell/fastcs2/blob/main/src/fastcs2/attribute_io_ref.py)
- Updated [Attribute*](https://github.com/GDYendell/fastcs2/blob/main/src/fastcs2/attribute.py)
- Updated [Controller](https://github.com/GDYendell/fastcs2/blob/main/src/fastcs2/controller.py)

jsouter:
Closes #211

all the tests are currently broken, have not fully removed handlers yet.
So far mostly copying https://github.com/GDYendell/fastcs2 with a few minor changes

- for the time being i've removed one of the generic args from AttributeIOs, I intend to recover them at some point
- using get_args to inspect the generic args for AttributeIO to prevent having to call `super().__init__(ref)`
- using TypeVar from typing_extensions which allows us to have a default value (this is nice for Attrs but might be problematic for AttributeIOs in the current implementation, will think about this a bit more)

very wip


src/fastcs/controller.py
```diff
@@ -98,6 +143,24 @@ class method and a controller instance, so that it can be called from any
...
 
+    def _validate_io(self, ios: Sequence[AttributeIO[T, AttributeIORefT]]):
+        """Validate that each Attribute has an AttributeIORef for which the
+        controller has an associated AttributeIO class, and that no two AttributeIO
+        classes handle the same AttributeIORef type"""
+        for ref_type, count in Counter([io.ref_type for io in ios]).items():
+            if count > 1:
+                raise RuntimeError(
+                    f"More than one AttributeIO class handles {ref_type.__name__}"
+                )
```
GDYendell, line 154: Ah, good one. I hadn't thought about this problem.

PR #223, opened 2025-09-30
Update transport model creation for new class structure

GDYendell:


GDYendell:
> should we rename the base `Transport` class to `BaseTransport`

Yeah it is something to consider. Though the reason `BaseController` exists currently is because `Controller` and `SubController` inherit it, so it is not quite the same. But, it might be a bit nicer if we have `BaseTransport` and then use`Transport` as the union of all transports for the model generation. I will leave it as is for now.

PR #225, opened 2025-10-02
Removes Backend class and moves logic and free functions to Launch fi…

### LINKED ISSUE #214: Split `Backend` into `FastCS` and `Controller`
The `FastCS` class partly superceded the `Backend` class, but some of the logic was left there. The distinction between what these two classes do is not obvious. [Here](https://github.com/GDYendell/fastcs2/blob/main/src/fastcs2/control_system.py#L12) a `Backend` class was not required and both `FastCS` and `Controller` do a bit more work.

## Acceptance Criteria
- `Backend` class is removed

gkalua:
closes #214


GDYendell:
Rebased on main and updated to work with recent changes.

GDYendell:
Thanks @gkalua !

PR #229, opened 2025-10-06
Fix softioc system test

### LINKED ISSUE #228: Fix startup of IOC system tests
tests failing on main, appears to be because we no longer `print(f"Running FastCS IOC: {self._pv_prefix}")` in the EPICS transport `.serve`s. 

`logger.info("Running IOC", pv_prefix=self._pv_prefix)` does not seem to get written to the `stdout_queue` in `run_ioc_as_subprocess` now

jsouter:
Closes #228

Small fix needed due to changes in logging messages that softioc system tests wait for before timing out. Also have removed src from the doctests for now as the Tracer docstring was being collected and failing under pytest.


GDYendell:
I think the coverage failure is not reliable because the tests are broken on main anyway.

PR #247, opened 2025-10-29
Implement Waveform

GDYendell:
Needs work in pvi to open plots on a new display and configure axis range

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,8 +30,8 @@ requires-python = ">=3.11"
 
 [project.optional-dependencies]
 demo = ["tickit~=0.4.3"]
-epicsca = ["pvi~=0.11.0", "softioc>=4.5.0"]
-epicspva = ["p4p", "pvi~=0.11.0"]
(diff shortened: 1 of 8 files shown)


jsouter:
https://github.com/epics-containers/pvi/pull/157

I was working on this in pvi this morning, just pushed onto my PR.
FastCS changes in scratch look like 
```
            case Waveform():
                if (length := len(fastcs_datatype.shape)) == 2:
                    default_width = 800
                    height = int(
                        default_width
                        * fastcs_datatype.shape[0]
                        / fastcs_datatype.shape[1]
                    )
                    return ImageRead(
                        width=default_width, height=height, color_map=ImageColorMap.GRAY
                    )
                elif length == 1:
                    return ArrayTrace(axis="")
                else:
                    raise NotImplementedError(
                        f"Waveform shape {fastcs_datatype.shape} not supported"
                    )
```
Also tries to add XY trace support for edl, though that's disabled in FastCS right now anyway.

PR #257, opened 2025-10-31
Make sure initial values are respected in CA transport

### LINKED ISSUE #202: initial_value not updated in transport if set() never called
I made an AttrRW with an Int() and an initial_value=1 and a handler that does nothing on put and update (AttrHandlerRW)
If I call .get() on the attr inside FastCS, the expected value of 1 is seen, but doing a caget on the equivalent PV (using the CA transport) the default value of 0 is seen, the callbacks don't get called in `__init__` and we need something to call attr.set() to trigger the callbacks. It's reasonable to not call all callbacks during `__init__` but we probably want the initial value to be set in the transport.

Of course this is a case where we can set the update_period to ONCE and set to the initial value in the update method of the handler, but this is a little convoluted. 

context: the old version of eiger 1 API lacks the "nexpi" parameter provided by the Eiger 2 API (or at least sometime between SIMPLON 1.6 and 1.8), I wanted to make a dummy attribute which just provides 1, and this default gets dynamically overwritten during initilisation/introspection if the parameter exists in the Eiger API, this parameter is needed for dodal/ophyd-async

jsouter:
Closes #202 

Proof of concept, it would be nicer if we could avoid the need for a system test/use of cagets as these can be flaky.

### DIFF
--- a/src/fastcs/transport/epics/ca/util.py
+++ b/src/fastcs/transport/epics/ca/util.py
@@ -1,4 +1,5 @@
 from dataclasses import asdict
+from typing import Any
 
 from softioc import builder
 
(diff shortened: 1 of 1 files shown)


GDYendell (commented):
I think this is good and testing through caget is reasonable. See if running fastcs.serve as a task (like [here](https://github.com/DiamondLightSource/FastCS/blob/9990065a96a27a2416eaa916c060c1dadfd12d2d/tests/test_control_system.py#L186)) works and makes it less flaky.

GDYendell:
This looks really neat! I don't mind it spying record set of doing caget, but it looks like the way you have done it is probably easier than caget?

jsouter:
Yeah I think it's easier this way. Will try and fix up the pyright issues

GDYendell (changes requested):

PR #284, opened 2025-12-09
Add AttrR wait_for_value

GDYendell:
Add `wait_for_value` to `AttrR` as a convenient way to wait for specific state before continuing with logic. 

This also makes `on_update` blocking by default for CA out records.


shihab-dls (approved):
Looks good! Approving!

PR #294, opened 2025-12-17
Fix command buttons in pva transport

GDYendell:
The pva transport expects the value for commands to be a boolean, while the ca transports expects it to be an int.


shihab-dls (approved):
Changes and tests look good! Tried it out with `example_p4p_ioc` as well. Approving!

PR #295, opened 2025-12-19
Windows: add guards for some methods that don't work/exist on windows

Tom-Willemsen:
This PR contains some minimal changes to allow a FastCS driver to start on Windows.

- `loop.add_signal_handler()` is not available on windows. On windows, signals are always delivered to the main thread of an application. Ctrl-c does correctly stop a FastCS driver in non-interactive mode on windows without this handler.
- `StdoutProxy` didn't work on windows under pytest - I believe it is conflicting with pytest's stdout capturing. Just use `sys.stdout` for that case.

FastCS' own unit tests don't pass on Windows even with this change - that would require a larger PR.

### DIFF
--- a/src/fastcs/logging/_logging.py
+++ b/src/fastcs/logging/_logging.py
@@ -33,18 +33,24 @@ def _configure_logger(
     logger.remove()
 
     try:
-        out = StdoutProxy(raw=True)
+        sink = StdoutProxy(raw=True)
     except Exception:
         # e.g. prompt_toolkit.output.win32.NoConsoleScreenBufferError on windows
         # But it isn't exported from prompt_toolkit in a cross-platform way.
-        out = sys.stdout
+        sink = sys.stdout
 
     logger.add(
-        sink=out,  # type: ignore
+        sink=sink,  # type: ignore
         colorize=True,
         format=format_record,
         level=level or "INFO",
     )
+
+    if sink == sys.stdout:
+        logger.warning(
+            "Unable to use prompt_toolkit.patch_stdout.StdoutProxy "
+            "as a logging sink; falling back to using sys.stdout"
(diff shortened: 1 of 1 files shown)


src/fastcs/logging/_logging.py
```diff
@@ -30,8 +31,9 @@ def _configure_logger(
     graylog_env_fields: GraylogEnvFields | None = None,
 ):
     logger.remove()
+    is_pytest_on_windows = "PYTEST_VERSION" in os.environ and os.name == "nt"
     logger.add(
-        sink=StdoutProxy(raw=True),  # type: ignore
+        sink=sys.stdout if is_pytest_on_windows else StdoutProxy(raw=True),  # type: ignore
```
GDYendell: Could this be moved to a fixture in the tests to keep the production code unchanged? I think something like this would do it

```python
@pytest.fixture(scope="session", autouse=True)
def patch_stdoutproxy_on_windows(mocker: MockerFixture):
    if os.name == "nt":
        import fastcs.logging._logging as logging_mod

        mocker.patch.object(
            logging_mod,
            "StdoutProxy",
            return_value=sys.stdout,
        )
```

src/fastcs/logging/_logging.py
```diff
@@ -30,8 +31,16 @@ def _configure_logger(
     graylog_env_fields: GraylogEnvFields | None = None,
 ):
     logger.remove()
+
+    try:
+        out = StdoutProxy(raw=True)
```
GDYendell:
```suggestion
        stdout = StdoutProxy(raw=True)
```

GDYendell (changes requested):
This looks good. Just a couple of minor changes.

PR #340, opened 2026-03-13
Simplify event loop handling and remove deprecated get_event_loop

GDYendell:
- [x] Merge #338 first

---

In the process of writing documentation it became clear that

- We are using get_event_loop, which is deprecated and planned for removal in 3.16
- The event loop handling is over complicated

Summary:
- Remove the `loop` parameter to `FastCS` and instead require clients to call `serve` from an async context with the loop they want it to run in
- Don't pass a `loop` to transports and instead retrieve it via `get_running_loop` only where required (EpicaCA and Tango)
- Use `asyncio.create_task` from an async context consistently in tests, rather than `run_until_complete` or `ensure_future`
- Improve tango test handling of DeviceTestContext to fix resource warnings

## Summary by CodeRabbit

* **Refactor**
  * Components no longer require supplying an explicit event loop; runtime loop is obtained internally, and task scheduling uses modern asyncio APIs.
* **Documentation**
  * Added a comprehensive guide covering startup, runtime, interactive shell, and shutdown flows.
* **Tests**
  * Updated tests to modern async patterns and to match the simplified startup/connect flow.

### DIFF
--- a/docs/explanations/control_system.md
+++ b/docs/explanations/control_system.md
@@ -1,0 +1,86 @@
+# The FastCS class
+
+`FastCS` is the entrypoint for a fastcs application. It connects a `Controller` to one
+or more `Transport`s, runs the controller's update loops as background tasks, and
+manages the full application lifecycle from startup to shutdown.
+
+## Construction
+
+```python
+from fastcs import FastCS
+from fastcs.controllers import Controller
+from fastcs.transports.epics import EpicsCATransport
+
+
+class MyController(Controller):
+    pass
+
+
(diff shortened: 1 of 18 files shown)


GDYendell:
@coderabbitai review

coretl (commented):
@ajgdls is now our resident Tango expert so I'll defer to him to review the tango bits. Everything else looks fine

GDYendell:
This is now simpler than it was initially and we only need per-test ignores on resource warnings.

PR #347, opened 2026-04-15
fix: add 'r' and 'w' entries for AttrRWs

### LINKED ISSUE #346: PVI structure does not include RBV
Currently, if we have an `AttrRW`, we add:
```python
p4p_raw_value[pv_leaf][attribute.access_mode] = pv
```
where `attribute.access_mode` is `rw`; thus, our served structure is:

```
structure a
    string rw P4P_TEST_DEVICE:A
```
However, we also do:

```python
case AttrRW():
    attribute_pv = make_shared_write_pv(attribute)
    attribute_pv_rbv = make_shared_read_pv(attribute)
    provider.add(f"{full_pv_name}", attribute_pv)
    provider.add(f"{full_pv_name}_RBV", attribute_pv_rbv)
```
so we end up serving an `_RBV`, but not including it into the pvi structure. On `ophyd-async`, this leads to the same `read_pv` and `write_pv`, and thus a `get_value()` on a `SignalRW` returns the setpoint value.

## Steps To Reproduce
1. Launch an example IOC on FastCS
2. Check PVI structure, then check generated gui
3. See served "_RBV" not included in PVI structure

## Acceptance Criteria
- If `AttrRW`, add entries for both `r` and `w` into the PVI structure

shihab-dls:
closes #346 

Amends PVI structure, such that we get:
```
structure a
  string r P4P_TEST_DEVICE:A_RBV
  string w P4P_TEST_DEVICE:A
```

## Summary by CodeRabbit

* **Enhancements**
  * Improved attribute access handling in EPICS PVA transport: read-write attributes now expose separate read and write fields.
  * Readback operations explicitly use `_RBV` suffix PVs for clearer value reporting.


tests/transports/epics/pva/test_p4p.py
```diff
@@ -33,11 +33,12 @@ async def test_ioc(p4p_subprocess: tuple[str, Queue]):
...
     assert parent_pvi["display"] == {"description": "some controller"}
     assert parent_pvi["value"] == {
-        "a": {"rw": f"{pv_prefix}:A"},
+        "a": {"w": f"{pv_prefix}:A", "r": f"{pv_prefix}:A_RBV"},
         "b": {"w": f"{pv_prefix}:B"},
         "child": {"d": f"{pv_prefix}:Child:PVI"},
         "table": {
-            "rw": f"{pv_prefix}:Table",
+            "r": f"{pv_prefix}:Table_RBV",
+            "w": f"{pv_prefix}:Table",
```
coretl, line 41: Hmm, this means we would have both a read and write table. We should discuss this...

coretl: Decided on a different tack:
- FastCS always makes separate R and W PVs
- We exposure them in the UI in the most compact way possible

This might mean some Phoebus level construct that merges the 2 PVs into one widget, or 2 buttons that launch the 2 widgets

coretl (approved):

PR #382, opened 2026-05-20
refactor(controllers): constructor-pass path, drop set_path()

### LINKED ISSUE #373: refactor(controllers): constructor-pass path, drop set_path()
## Proposal (2d)

Make `path` a parameter of `Controller.__init__` and thread it through user code. Remove `set_path()` entirely. The launcher constructs the root with `cls(path=[entry.id], **options)`; a parent constructs its subs with the full path baked in:

```python
class Temp(Controller):
    def __init__(self, path: list[str], settings: TempSettings):
        super().__init__(path=path)
        self.add_sub_controller("ramp_0", Ramp(path=path + ["ramp_0"], ...))
```

`add_sub_controller(name, sub)` no longer mutates `sub._path` — it sanity-asserts `sub.path[-1] == name` (or just trusts the parent).

Result: one concept (`path`), no post-construction setters, no rootless-sub state, no two-phase init, no recursive re-prefix.

## Cost

Any custom `Controller.__init__` (root **and** sub-controllers) must accept `path` and forward to super:

```python
def __init__(self, path: list[str], my_settings: MySettings):
    super().__init__(path=path)
```

This is the same ergonomic shape as the rejected branch B (`yaml-list-id-positional`), applied to *every* customised `Controller`, not just roots. We're accepting it here because **`path` is a real Controller concept**, not a YAML-leaked one — user code already touches `self.path + [name]` when adding subs, so threading `path` through `__init__` is paying the cost where the concept naturally lives. That distinguishes it from `id`, which was a YAML label that shouldn't have leaked into user signatures (which is what motivated branch C).

The contrarian read: "2d is just branch B with `path` instead of `id`" — mechanically true. Be honest about that.

## Open sub-decisions

- `path` positional vs keyword-only on `Controller.__init__`. Lean kwonly (matches branch A) to avoid the sub-controller call-site fallout that branch B hit (`description=` had to become a kwarg to dodge the new positional slot).
- Does `path` default to `[]` for direct/standalone construction (tests, embedded use)? Probably yes.
- `add_sub_controller`'s residual role: pure registration + sanity-assert, or drop it and have parents just write `self.sub_controllers[name] = ...`?

## Context: how we got here

Branch C (landed in PR #360 as [`c8adfee7`](https://github.com/DiamondLightSource/fastcs/pull/360/commits/c8adfee7)) drops `_id` / `set_id` / `Controller.id`. The launcher does `c = cls(options); c.set_path([entry.id])`. The single production reader of `.id` (`control_system.py:26`, IPython context-key + dup-key validation) reads `path[0]` instead. YAML `id:` becomes purely a label for the root path segment.

Branch C deliberately leaves one wrinkle alone: when a root Controller adds subs in its `__init__`, those subs' `_path` is set to `[name]` (rootless), because the launcher hasn't seeded the root's `_path` yet. Today's `_build_api` papers over this by recursing with `path + [name]` from the root's seeded path, so emitted `ControllerAPI.path`s come out correct — but `sub_controller._path` itself remains rootless. That latent oddity exists on `multiple-controllers` today and has never bitten anyone; bundling its fix into branch C would obscure whether the id-removal itself was the right call. Hence this issue.

## Alternatives considered for fixing the rootless-sub state

- **2a — Two-phase Controller lifecycle.** Split sub-registration out of `__init__` into a framework-called `register_subs()` hook that runs after `set_path`. Forces every existing controller to migrate `add_sub_controller` calls out of `__init__`. Rejected: high churn for a problem users don't currently feel.
- **2b — Declarative sub-controllers.** Class-level descriptors / spec returned from `__init__`, framework instantiates with full path baked in. Rejected: framework machinery, awkward for runtime-determined subs (e.g. count from settings).
- **2c — Recursive re-prefix on `set_path`.** Make `set_path` walk descendants and re-prefix their `_path`s. Local change, invisible to users, preserves current authoring style. Plausible but keeps a post-construction setter that 2d eliminates entirely.
- **2d (this proposal).** Constructor-pass `path`. Drops the setter; subs constructed under a populated `self.path` are correct first time; no recursion needed.

2d wins because it's the only option that removes the post-construction setter and the rootless-sub state simultaneously. The cost (user `__init__` boilerplate for `path`) is the price.

gilesknap:
## Summary
Implements branch 2d from #373:
- `path` becomes a keyword-only constructor parameter on `BaseController` / `Controller` / `ControllerVector`; the launcher passes `path=[entry.id]` rather than calling `set_path` after construction.
- `BaseController.set_path()` is removed. `add_sub_controller(name, sub)` now sanity-asserts `sub.path == parent.path + [name]` instead of mutating `sub._path`. Parents construct subs with the full path baked in (e.g. `Sub(path=self.path + [name])`).
- `_build_api()` reads `self._path` directly; the parent-side `path` argument and recursive re-prefix are gone.
- Demo controllers, doc snippets and every test that previously did `controller.set_path([id])` are updated to thread `path=[...]` through `__init__`.

Closes #373.

## Notes on the open sub-decisions
- **Positional vs keyword-only**: `path` is keyword-only — keeps existing positional call-sites (description, ios, settings, …) intact.
- **Default**: `path` defaults to `[]` so direct/standalone construction (tests, embedded use) still works.
- **`add_sub_controller`'s residual role**: strict sanity-assert (full path match), not just `path[-1] == name`. This catches the common mistake of forgetting to thread `path` end-to-end and is what every updated test relies on for diagnostics.

## Test plan
- [x] `pytest tests/test_controllers.py tests/test_multi_controller.py tests/test_launch.py tests/test_attributes.py tests/test_methods.py tests/test_attribute_logging.py tests/test_control_system.py tests/test_util.py tests/test_ip_connection.py tests/transports/rest tests/transports/graphQL tests/transports/epics/test_emission.py tests/transports/epics/ca/test_gui.py tests/transports/epics/ca/test_softioc.py tests/transports/epics/ca/test_ca_util.py` — all pass locally (one pre-existing flaky thread-cleanup failure in `test_scan_exception_sets_disconnected_and_reconnect_resumes`, order-dependent, passes when run alone).
- [x] All doc snippets `static04..15` and `dynamic.py` import and execute cleanly.
- [ ] CI: full suite including p4p / tango / softioc / docs-snippets (need a real env).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Refactor**
  * Controllers now accept an optional path at construction; the prior separate path-setting step has been removed. Controller APIs now derive paths from each controller’s stored path.

* **Documentation**
  * Examples and snippets updated to pass path during controller construction instead of configuring paths afterward.

* **Tests**
  * Test suites and fixtures updated to construct controllers with explicit paths at instantiation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/DiamondLightSource/fastcs/pull/382?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/snippets/dynamic.py
+++ b/docs/snippets/dynamic.py
@@ -98,9 +98,11 @@ class TemperatureRampController(Controller):
         index: int,
         parameters: dict[str, TemperatureControllerParameter],
         io: TemperatureControllerAttributeIO,
+        *,
+        path=None,
     ):
         self._parameters = parameters
-        super().__init__(f"Ramp{index}", ios=[io])
+        super().__init__(f"Ramp{index}", ios=[io], path=path)
 
     async def initialise(self):
         for name, attribute in create_attributes(self._parameters).items():
@@ -108,12 +110,12 @@ class TemperatureRampController(Controller):
 
 
 class TemperatureController(Controller):
-    def __init__(self, settings: IPConnectionSettings):
+    def __init__(self, settings: IPConnectionSettings, *, path=None):
         self._ip_settings = settings
         self._connection = IPConnection()
 
         self._io = TemperatureControllerAttributeIO(self._connection)
-        super().__init__(ios=[self._io])
+        super().__init__(ios=[self._io], path=path)
 
     async def connect(self):
         await self._connection.connect(self._ip_settings)
@@ -129,19 +131,19 @@ class TemperatureController(Controller):
             self.add_attribute(name, attribute)
 
         for idx, ramp_parameters in enumerate(ramps_api):
+            name = f"Ramp{idx + 1:02d}"
             ramp_controller = TemperatureRampController(
-                idx + 1, ramp_parameters, self._io
+                idx + 1, ramp_parameters, self._io, path=self.path + [name]
             )
             await ramp_controller.initialise()
-            self.add_sub_controller(f"Ramp{idx + 1:02d}", ramp_controller)
+            self.add_sub_controller(name, ramp_controller)
 
         await self._connection.close()
 
 
 epics_ca = EpicsCATransport()
 connection_settings = IPConnectionSettings("localhost", 25565)
-controller = TemperatureController(connection_settings)
-controller.set_path(["DEMO"])
+controller = TemperatureController(connection_settings, path=["DEMO"])
 fastcs = FastCS(controller, [epics_ca])
 
 
--- a/docs/snippets/static04.py
+++ b/docs/snippets/static04.py
@@ -10,8 +10,7 @@ class TemperatureController(Controller):
 
 
 epics_ca = EpicsCATransport()
-controller = TemperatureController()
-controller.set_path(["DEMO"])
+controller = TemperatureController(path=["DEMO"])
 fastcs = FastCS(controller, [epics_ca])
 
 if __name__ == "__main__":
--- a/docs/snippets/static05.py
+++ b/docs/snippets/static05.py
@@ -14,8 +14,7 @@ class TemperatureController(Controller):
 
 gui_options = EpicsGUIOptions(output_dir=Path("."), title="Demo Temperature Controller")
 epics_ca = EpicsCATransport(gui=gui_options)
-controller = TemperatureController()
-controller.set_path(["DEMO"])
+controller = TemperatureController(path=["DEMO"])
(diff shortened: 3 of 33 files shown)


gilesknap:
@ggayDiamond where did we get to with this? If you want me to spend some time on it this week I will have some time to do that.

PR #385, opened 2026-05-28
feat: allow aliases to be passed to ca transport

shihab-dls:
Closes #384

This PR allows passing aliases to the `ca` transport. An Example would be:

```yaml
transport:
  - epicsca:
      aliases:
         - PV_TO_ALIAS: "ALIAS"
```

## Summary by CodeRabbit

* **New Features**
  * Added EPICS CA alias support via a new `aliases` transport option, registering alternate PV names for attributes, command PVs, and readback (`*_RBV`) variants.
  * Aliases are validated to prevent duplicate alias values; overly long alias names are skipped with a warning.
* **Tests**
  * Added coverage for alias registration for write, readback (`_RBV`), and command PVs, plus updated IOC initialization and expectations.
* **Documentation**
  * Updated the published EPICS CA options/schema to include the new `aliases` field.

### DIFF
--- a/src/fastcs/transports/epics/ca/ioc.py
+++ b/src/fastcs/transports/epics/ca/ioc.py
@@ -1,4 +1,5 @@
 import asyncio
+from collections import Counter
 from typing import Any, Literal
 
 from softioc import builder, softioc
@@ -29,6 +30,14 @@ class EpicsCAIOC:
     """A softioc which handles one or more controllers."""
 
     def __init__(self, controller_apis: list[ControllerAPI], aliases: dict[str, str]):
+        if duplicate_aliases := [
+            alias for alias, count in Counter(aliases.values()).items() if count > 1
+        ]:
+            raise RuntimeError(
+                "Failed to create EPICS CA IOC, as duplicate aliases were provided:"
+                f" {duplicate_aliases}"
+            )
+
         self._controller_apis = controller_apis
         for controller_api in controller_apis:
             root_pv_prefix = pv_prefix_from_path(controller_api.path)
@@ -319,4 +328,10 @@ def _add_attr_pvi_info(
 
 def _add_alias(record: RecordWrapper, alias: str | None):
     if alias is not None:
-        record.add_alias(alias)
+        if len(alias) > EPICS_MAX_NAME_LENGTH:
+            logger.warning(
+                f"Not creating alias {alias}, as full name would exceed"
+                f" {EPICS_MAX_NAME_LENGTH} characters"
+            )
+        else:
+            record.add_alias(alias)
--- a/src/fastcs/transports/epics/ca/ioc.py
+++ b/src/fastcs/transports/epics/ca/ioc.py
@@ -34,7 +34,7 @@ class EpicsCAIOC:
             _add_sub_controller_pvi_info(controller_api)
 
             _create_and_link_attribute_pvs(controller_api, aliases)
-            _create_and_link_command_pvs(controller_api)
+            _create_and_link_command_pvs(controller_api, aliases)
 
     def run(
         self,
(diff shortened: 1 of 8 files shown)


coretl (commented):
Implementation looks fine.

Are there any system tests for CA? If so, then it should be added there too.

src/fastcs/transports/epics/ca/ioc.py
```diff
@@ -168,13 +183,22 @@ async def async_record_set(value: DType_T):
         record.set(cast_to_epics_type(attribute.datatype, value))
 
     record = _make_in_record(pv, attribute)
+
+    if alias:
+        suffix = "_RBV" if pv_name.endswith("_RBV") else ""
```
gilesknap: I think you are following previous convention here, but _RBV should be a string constant declared once.

gilesknap: But do we want to repeat _RBV handling? Do you specifically want to only override _RBV sometimes, or could we just specify the root name and alias both always. Probably either is OK but this behaviour should be described somewhere.

gilesknap (commented):
Looks good. One minor comment inline, plus this is for CA only. Won't PVA want it too?

src/fastcs/transports/epics/ca/ioc.py
shihab-dls: That's a good point; Sometimes, a controller `AttrR` has to be aliased to a legacy pv with an `_RBV` suffix, which makes handling "_RBV" aliases internally for `AttrRW`s odd. I'll remove this handling and define an RBV constant at the top.

shihab-dls:
> Looks good. One minor comment inline, plus this is for CA only. Won't PVA want it too?

I think @coretl and I agreed that, for now at least, CA is the only transport that required aliasing, as we're doing this for backwards compatibility with GDA.

gilesknap:
@coderabbitai full review

gilesknap:
@shihab-dls I feel that the _RBV is still possibly confusing:

1. its predicated on the type of the attrubute but you don't specify that in the config - only the name
2. RW assume that they can add _RBV (for the readback) and alias the demand and readback
3. R only takes the name verbatim which I take to mean _RBV is included

I can't remember - if the RW attribute readback name is not a simple _RBV suffix, does that mean the attributes need to be declared as separate R and W? If not then 2 would break. Either way their are some assumptions here that don't immediately jump out.

Maybe we don't care?

shihab-dls:
> @shihab-dls I feel that the _RBV is still possibly confusing:
> 
>     1. its predicated on the type of the attrubute but you don't specify that in the config - only the name
> 
>     2. RW assume that they can add _RBV (for the readback) and alias the demand and readback
> 
>     3. R only takes the name verbatim which I take to mean _RBV is included
> 
> 
> I can't remember - if the RW attribute readback name is not a simple _RBV suffix, does that mean the attributes need to be declared as separate R and W? If not then 2 would break. Either way their are some assumptions here that don't immediately jump out.
> 
> Maybe we don't care?

It is indeed the case that, if you need an AttrRW with different PV names for the setpoint and readback, then you need to make a separate AttrW and AttrR; making an AttrRW currently implies that you intend to have a setpoint and an `_RBV`. This does seem quite lacking, and maybe we need readback PV overrides on AttrRWs. I'll make an issue to address this, but for this PRs scope I think we can just work with this implication in mind.

There is a case where we want to alias:
```
ID:PV: "ID:SomeAlias"
ID:PV_RBV: "ID:AnotherAlias"
```
where our **aliased** PVs need to be Different for AttrRWs. This is likely, as legacy IOCs can have odd PV naming. Because of this, this [comment](https://github.com/DiamondLightSource/fastcs/pull/385#discussion_r3371568036) of yours is very valid, and we shouldn't handle making the alias `_RBV`s ourselves. This is why the current implementation checks for an alias assigned to a readback PV, instead of making the readback alias itself. This way, a list of aliases you provide must be complete, and no additional aliases are created internally.

EDIT: Either way, I should document this somewhere in the fastcs docs. I'll add that to the new issue I'll create about AttrRWs, just so it doesn't block this PR.

gilesknap (approved):
That seems fine to me. Approved.

PR #389, opened 2026-07-14
fix: Raise Exceptions on Hardware I/O to Set Record Alarms

### LINKED ISSUE #303: Implement `op.done(error="msg")` for failed put operations on command and write PVs
We have found that doing `op.done(error="msg")` for a `command` results in a pop up dialogue box, but doing the same for `WritePv`s does not. Both cases result in a log to the `phoebus` command line of the error message. We should investigate why this occurs, and check if we can prevent the pop up for commands, such that we can implement `op.done(error="msg")` appropriately.

## Acceptance Criteria
- Specific criteria that will be used to judge if the issue is fixed

shihab-dls:
Closes #303

This depends on a new release of `softioc`.

This PR sets and resets alarm states on failed and successful PV puts for CA and PVA.

This addresses the underlying bug in [ophyd-async](https://github.com/bluesky/ophyd-async/issues/1315), where a failed command does not raise an error. However, only PVA allows these errors to immediately propagate up to the signal call, otherwise requiring a check on the PV status after the put on CA. Therefore, fastcs-eiger/odin must be converted to use PVA to close the ophyd-async bug.

## Summary

* **Bug Fixes**
  * Attribute put failures and command executions now properly propagate errors instead of being silently ignored.
  * EPICS Channel Access and PVAccess now clear alarms on success and set major alarms on failures.
  * Failed command/put error messages are more consistent and include exception details.
  * Improved controller sub-item PV key casing for generated raw values.
* **Packaging / Chores**
  * Updated formatting and the source used for an optional EPICS-related extra install.
* **Tests**
  * Added and expanded coverage for alarm behavior and error propagation across transports.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ requires-python = ">=3.11"
 demo = ["tickit~=0.4.3"]
 epicsca = [
     "pvi~=0.12.0",
-    "softioc @ git+https://github.com/DiamondLightSource/pythonSoftIOC.git@master",
+    "softioc>=4.7.1",
 ]
 epicspva = ["p4p", "pvi~=0.12.0"]
 epics = ["fastcs[epicsca]", "fastcs[epicspva]"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,8 +5,8 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "fastcs"
 authors = [
-    {name = "Martin Gaughran", email = "martin.gaughran@diamond.ac.uk"},
-    {name = "Gary Yendell", email = "gary.yendell@diamond.ac.uk"},
+    { name = "Martin Gaughran", email = "martin.gaughran@diamond.ac.uk" },
+    { name = "Gary Yendell", email = "gary.yendell@diamond.ac.uk" },
 ]
 classifiers = [
     "Development Status :: 3 - Alpha",
@@ -34,7 +34,10 @@ requires-python = ">=3.11"
 
 [project.optional-dependencies]
 demo = ["tickit~=0.4.3"]
-epicsca = ["pvi~=0.12.0", "softioc>=4.5.0"]
+epicsca = [
+    "pvi~=0.12.0",
+    "softioc @ git+https://github.com/DiamondLightSource/pythonSoftIOC.git@master",
+]
 epicspva = ["p4p", "pvi~=0.12.0"]
 epics = ["fastcs[epicsca]", "fastcs[epicspva]"]
 tango = ["pytango"]
--- a/src/fastcs/transports/epics/ca/ioc.py
+++ b/src/fastcs/transports/epics/ca/ioc.py
@@ -230,13 +230,9 @@ def _create_and_link_write_pv(
         logger.info("PV put: {pv} = {value}", pv=pv, value=repr(value))
         try:
             await attribute.put(cast_from_epics_type(attribute.datatype, value))
+            _set_alarm(record, alarm.NO_ALARM)
         except Exception:
-            record.set(
-                record.get(),
-                process=False,
-                severity=alarm.MAJOR_ALARM,
-                alarm=alarm.MAJOR_ALARM,
-            )
+            _set_alarm(record, alarm.MAJOR_ALARM)
 
     async def set_setpoint_without_process(value: DType_T):
         tracer.log_event(
@@ -290,13 +286,9 @@ def _create_and_link_command_pv(
 
         try:
             await method.fn()
+            _set_alarm(record, alarm.NO_ALARM)
         except Exception:
-            record.set(
-                record.get(),
-                process=False,
-                severity=alarm.MAJOR_ALARM,
-                alarm=alarm.MAJOR_ALARM,
-            )
+            _set_alarm(record, alarm.MAJOR_ALARM)
 
     record = builder.Action(
         f"{pv_prefix}:{pv_name}",
@@ -350,3 +342,12 @@ def _add_alias(record: RecordWrapper, alias: str | None):
             )
         else:
             record.add_alias(alias)
+
+
+def _set_alarm(record: RecordWrapper, alarm_state: int):
+    record.set(
+        record.get(),
+        process=False,
+        severity=alarm_state,
+        alarm=alarm_state,
+    )
--- a/src/fastcs/transports/epics/ca/ioc.py
+++ b/src/fastcs/transports/epics/ca/ioc.py
@@ -1,7 +1,6 @@
 import asyncio
 from collections import Counter
 from collections.abc import Awaitable
-from enum import IntEnum
 from typing import Any, Literal
 
 from softioc import alarm, builder, softioc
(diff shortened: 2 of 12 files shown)


src/fastcs/transports/epics/pva/_pv_handlers.py
```diff
@@ -48,13 +50,28 @@ async def put(self, pv: SharedPV, op: ServerOperation):
...
 
-        await self._attr_w.put(cast_value)
-        op.done()
+        value_to_post = (
+            cast_to_p4p_value(self._attr_w, cast_value)
+            if isinstance(datatype, Waveform | Table | Enum)
+            else cast_value
+        )
+
+        _post_with_alarm_states(pv, datatype, value_to_post, p4p_alarm_states())
```
olliesilvester: Not 100% sure but can we just do


```suggestion
        _post_with_alarm_states(pv, datatype, raw_value, p4p_alarm_states())
```

olliesilvester: Also, as nit and out of scope, but `p4p_alarm_states` could return a validated pydantic basemodel instead of `dict`

src/fastcs/transports/epics/ca/ioc.py
```diff
@@ -218,18 +218,23 @@ async def async_record_set(value: DType_T):
...
-    pv_name: str,
-    attr_name: str,
-    alias: str | None,
-    attribute: AttrW[DType_T],
-) -> None:
+    pv_prefix,
+    pv_name,
+    attr_name,
+    alias,
+    attribute,
```
olliesilvester: How come the typing got removed?

src/fastcs/transports/epics/ca/ioc.py
```diff
@@ -218,18 +218,23 @@ async def async_record_set(value: DType_T):
...
         logger.info("PV put: {pv} = {value}", pv=pv, value=repr(value))
-
-        await attribute.put(cast_from_epics_type(attribute.datatype, value))
+        try:
+            await attribute.put(cast_from_epics_type(attribute.datatype, value))
+            _set_alarm(record, alarm.NO_ALARM)
+        except Exception:
+            # No need to log, as AttrW.put already logs exceptions
+            # So just need to set alarm
+            _set_alarm(record, alarm.MAJOR_ALARM)
```
olliesilvester: Nit: This little try except pattern is only used twice, but could pull into a helper decorator / wrapper

src/fastcs/transports/epics/ca/ioc.py
```diff
@@ -335,3 +346,12 @@ def _add_alias(record: RecordWrapper, alias: str | None):
             )
         else:
             record.add_alias(alias)
+
+
+def _set_alarm(record: RecordWrapper, alarm_state: int):
```
olliesilvester, line 341: nit: I'd make `alarm_state` an `IntEnum` type so the variable names can describe what they mean

src/fastcs/transports/epics/pva/_pv_handlers.py
```diff
@@ -165,3 +187,15 @@ def _wrap(value: dict):
     )
 
     return shared_pv
+
+
+def _post_with_alarm_states(
```
olliesilvester, line 192: Could we make this a public function? Along with the `BaseModel` for `alarm_states` this would be very useful

olliesilvester (commented):
Not finished the review just yet but some initial comments

tests/transports/epics/pva/test_p4p.py
```diff
@@ -148,7 +149,9 @@ async def test_command_method(p4p_subprocess: tuple[str, Queue]):
         assert after_command_value["value"] is False
         assert after_command_value["alarm"]["severity"] == 2
         assert (
-            after_command_value["alarm"]["message"] == "I: FAILED WITH THIS WEIRD ERROR"
+            after_command_value["alarm"]["message"]
+            == "Exception raised during command put: "
+            "RuntimeError('I: FAILED WITH THIS WEIRD ERROR')"
```
olliesilvester, line 154: nit: could use constants here instead of copying strings

tests/transports/epics/pva/test_p4p.py
```diff
@@ -209,6 +212,37 @@ async def test_numeric_alarms(p4p_subprocess: tuple[str, Queue]):
...
+
+    class SomeController(Controller):
+        a = AttrW(Int(), io_ref=SimpleAttributeIORef())
+
+    controller = SomeController(ios=[SimpleAttributeIO()])
+    pv_prefix = str(uuid4())
+    fastcs = make_fastcs(pv_prefix, controller)
+
+    async def put_pvs():
+        await asyncio.sleep(0.1)
```
olliesilvester:
```suggestion
        await asyncio.sleep(0.01)
```
or can we get away with `asyncio.sleep(0)`

tests/transports/epics/pva/test_p4p.py
```diff
@@ -209,6 +212,37 @@ async def test_numeric_alarms(p4p_subprocess: tuple[str, Queue]):
         a_monitor.close()
 
 
+def test_alarms_set_if_put_fails(caplog):
```
olliesilvester: I don't think caplog is being used

tests/transports/epics/pva/test_p4p.py
```diff
@@ -209,6 +212,37 @@ async def test_numeric_alarms(p4p_subprocess: tuple[str, Queue]):
...
+        await ctxt.put(f"{pv_prefix}:A", 1)
+
+    serve = asyncio.ensure_future(fastcs.serve(interactive=False))
+    try:
+        asyncio.get_event_loop().run_until_complete(
+            asyncio.wait_for(asyncio.gather(serve, put_pvs()), timeout=1)
+        )
+    except RuntimeError as exc:
+        assert str(exc) == "Exception raised during put operation: ValueError('Failed')"
+    serve.cancel()
```
olliesilvester: Better to do a `with pytest.raises` here probably. Right now if a `RuntimeError` isn't raised, the test will pass

tests/example_softioc.py
```diff
@@ -18,11 +19,19 @@ class ParentController(Controller):
...
 
 class ChildController(Controller):
+    fail_on_next_e: bool = True
     c: AttrW = AttrW(Int())
 
     @command()
     async def d(self):
-        pass
+        print("D: RUNNING")
+        await asyncio.sleep(0.1)
```
olliesilvester: Similar to above, can we use a shorter sleep time, or 0 ?

pyproject.toml
```diff
@@ -34,7 +34,10 @@ requires-python = ">=3.11"
 
 [project.optional-dependencies]
 demo = ["tickit~=0.4.3"]
-epicsca = ["pvi~=0.12.0", "softioc>=4.5.0"]
+epicsca = [
+    "pvi~=0.12.0",
+    "softioc @ git+https://github.com/DiamondLightSource/pythonSoftIOC.git@master",
+]
```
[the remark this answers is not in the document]
olliesilvester: Might be worth adding a comment / issue to remind to pin this back to a version on next release

olliesilvester (commented):
I don't have permission to actually approve anymore, but I would have

Left a bunch of comments but the fix works, thanks!

shihab-dls:
@coretl Silvester doesn't have approval permission, so I'm requesting you as a second reviewer.

coretl (approved):

PR #402, opened 2026-07-20
WIP: What is the current state of the refactor

### LINKED ISSUE #388: ophyd-async / FastCS API Convergence
## Summary

FastCS and ophyd-async have a lot of conceptual similarities: type hint friendly ways of making a hardware abstraction library. There is also a lot of crossover of developers from one to another, so we should make sure that similar concepts are done in the same way, and there are no "false friends" that could mislead people into making a connection that is not actually there. This report details the current state of play, and makes suggestions for how FastCS could change to be closer to ophyd-async as they both approach 1.0 releases. It was AI generated from local checkouts of both modules, then initially iterated on by myself. It is posted here for wider comment.

## 1. Background — the two APIs today

| Concept | ophyd-async | FastCS |
|---|---|---|
| Tree node | `Device` (+ pluggable `DeviceConnector`) | `BaseController` / `Controller` |
| Leaf value | `SignalR/W/RW[T]` over `SignalBackend` | `AttrR/W/RW[T, RefT]` |
| Action | `Command[P, T]`, `TriggerableCommand` | `@command()` — void/void only |
| Int-keyed group | `DeviceVector[T]` | `ControllerVector[T]` (already near-identical) |
| Declarative | class-body hints **create** children (`DeviceFiller`); metadata in `Annotated` extras (`PvSuffix`, `TangoPolling`, `Format`); *filled* now or *unfilled* until connect-time introspection (`PviDeviceConnector`, `TangoDeviceConnector`) | class-body `AttrRW(Int(), io_ref=...)` **instances** deep-copied per instance (`_bind_attrs`); hints only *validate* (`HintedAttribute`) |
| Procedural | `soft_signal_rw(...)`, `soft_command(cb)` in `__init__` | `add_attribute`/setattr in `initialise()`; IO via `AttributeIO` classes dispatched by `AttributeIORef` type |
| Value types | plain types in generics: `SignalRW[float]`, `Array1D[np.int32]`, `Table` | `DataType` instances: `Float(prec=3)`, `Waveform(np.int32, (4,))`, `Enum(cls)` |
| Serving | n/a (client) | `FastCS(controllers, transports)` → CA/PVA(+PVI), Tango, REST, GraphQL |

Existing bridge: `ophyd_async.fastcs.core.fastcs_connector(uri)` =
`PviDeviceConnector` (PVA+PVI over the network; PVI `"x"` entries already map
to `TriggerableCommand`).

---

## 2. Decisions

Agreed with Giles (maintainer of both):

1. **Shared declarative idiom = the annotation mechanism** (hints create
   children; a filler provisions them at init or connect). Rule stated
   identically in both projects: **class body = declarations, instance scope
   = procedural construction**. FastCS class-scope Attribute prototypes and
   the deepcopy half of `_bind_attrs` are removed pre-1.0. Rationale (§3.1):
   the instance style must deepcopy (aliasing hazard), cannot express
   "declared now, filled by introspection at connect" (why `HintedAttribute`
   exists as a second mechanism), and ophyd-async *cannot* move in the other
   direction because a Signal's backend depends on which connector the
   instance is created with.
2. **Harsh split in FastCS**: declarative = bare hints only, made concrete at
   connect/initialise time by introspection via a new `ControllerFiller`.
   All IO/function attachment is procedural, in `__init__`. **The
   `AttributeIORef` / IO-registration machinery is removed** (§3.2).
3. **The generic extras mechanism stays**: `ControllerFiller` yields
   `(child, extras)` like `DeviceFiller` does, so protocol libraries (e.g. a
   future SCPI package) can define their own `Annotated` extras. Core FastCS
   defines **no** extras vocabulary for 1.0.
4. **AttributeIO survives as a per-attribute, connection-bound object**
   passed as a single `io=` argument, in **R / W / RW flavours with abstract
   `update` / `send` methods** so access-mode compatibility is checked
   statically (§3.3). Verbs stay `update`/`send` (not getter/setter).
5. **FastCS gains typed `Command[P, T]` semantics for 1.0** (args + return).
   EPICS transports stay void/void; Tango/REST/GraphQL and the embedded
   connector serve typed commands fully.
6. **Embedding = ophyd-async-side `DeviceConnector`** consuming a **stable
   public Controller interface** (§5), packaged behind an
   **`ophyd-async[fastcs-embed]` extra**. No shared package; convergence by
   convention.
7. **`DataType` classes stay** as the procedural/runtime/introspection value
   (`AttrRW(Float(prec=3, units="K"))`). Classmethod factories
   (`AttrRW.float(...)`) rejected: classmethod return types can't be
   inherited generically (no `Self[float]`), forcing ~24 duplicated
   signatures, and they produce no value usable in `Annotated` slots or by
   introspection code. The `Float`/`float` declarative duplication dissolves
   under decision 2 (bare hints; metadata comes from introspection); a
   `Prec`/`Units`/`Shape` extras vocabulary is a post-1.0 option via the
   extras mechanism.

Also agreed (answers in `prompt.md` round 3):

8. **Embedded lifecycle**: connector owns a runner (started on first
   `connect_real`, idempotent); shutdown via an **atexit hook** (Giles:
   "atexit is fine") plus explicit `await connector.shutdown()`;
   `Device.disconnect()` upstream as a follow-up, don't block on it.
9. **Embedded + transports simultaneously** (e.g. CA GUI next to bluesky):
   yes, but out of scope for the first cut; design the runner so a
   transport list can be added later.
10. **FastCS gaps — both for 1.0**: (a) `AttrW` caches its setpoint (needed
    for ophyd `locate()`; useful to all transports); (b) **FastCS-native
    timestamps** (and severity where meaningful) on attribute updates —
    Giles: useful to all transports (Tango event pushes, EPICS records),
    not just the connector. Connector stamps receive-time only as an
    interim until (b) lands.
11. **Typed commands over EPICS transports**: skip-with-warning (serve the
    void ones, warn about the typed ones). Hard error only where the user
    *declares* a typed command on something EPICS-only (matches ophyd-async
    connector behaviour).
12. **Naming alignment** (while breaking pre-1.0): `prec` → `precision`;
    align `min/max/min_alarm/max_alarm` with event-model `Limits` naming;
    adopt `Array1D[np.int32]` / `Table` as FastCS *hint* spellings (mapping
    to `Waveform`/table `DataType`s internally).
13. **Stable interface home**: formalise in fastcs core — `ControllerAPI`,
    the attribute/command runtime methods, and a `ControllerRunner` (or
    `Controller.serve()/stop()`) extracted from `FastCS.serve`. The ophyd
    connector imports only that documented surface.
14. **Attr-from-method decorator sugar** (from the PyTango comparison,
    §7.5): property-style decorators (`@attr_r` / `@attr_rw` + `.send`)
    that create an Attribute bound to controller methods, so the trivial
    case is one decorated getter. Pure sugar over `AttrR/W/RW` + callback
    IO, built on the existing `UnboundCommand`/`UnboundScan` bind machinery
    (fresh objects per instance — no prototype/deepcopy hazard). Refines
    the class-body rule to: *class body = declarations + decorated
    behaviour; instance scope = construction with data* (already true today
    via `@command`/`@scan`).

---

### 3.1 Why annotations won (summary of the analysis)

- **Prototype aliasing**: class-scope instances must deepcopy or all
  controller instances share one mutable Attribute; deepcopy is fragile
  (connections/locks/bound callbacks) and costs construction time.
  Annotations are inert — nothing to copy.
- **Declared-but-filled-later**: instances must be complete at class
  definition; annotations natively express "exists, typed, provisioned by
  introspection at connect" (`filled=False` + `check_filled`), unifying
  FastCS's two mechanisms (instances + `HintedAttribute`) into one.
- **Typing**: hint == runtime by construction; extras invisible to checkers.
- **The asymmetry**: ophyd-async can't adopt instance style (backend depends
  on connector choice + connect-time data), so one shared idiom is only
  possible in this direction.
- **Ecosystem**: same idiom as pydantic v2 `Annotated[int, Field(...)]`.

### 3.2 Remove `AttributeIORef` / IO registration

Verified: **nothing downstream consumes `io_ref` or the IO registry** — no
transport touches them; `_connect_attribute_ios` bottoms out in
`attr.set_on_put_callback(io.send)` / `attr.set_update_callback(io.update)`,
i.e. the split is a dispatch layer over callbacks that already exist. Its
sole structural justification was that class-scope Attributes are created
before `__init__`, so they couldn't close over a connection. Decision 2
removes that situation entirely.

Delete: `AttributeIORef`, ref-type dispatch (`__init_subclass__` generic
sniffing), `ios=` kwarg, `_validate_io`, `_connect_attribute_ios`,
`_attribute_ref_io_map`, and the **second TypeVar** — `Attribute[DType_T,
AttributeIORefT]` collapses to `Attribute[DType_T]`, making `AttrRW[float]`
structurally isomorphic to `SignalRW[float]`.

### 3.3 New `AttributeIO`: per-attribute, R/W/RW flavours, single `io=` arg

```python
class ReadIO(Generic[DType_T], ABC):
    def __init__(self, update_period: float | None = None): ...
    @abstractmethod
    async def update(self, attr: AttrR[DType_T]) -> None: ...

class WriteIO(Generic[DType_T], ABC):
    @abstractmethod
    async def send(self, attr: AttrW[DType_T], value: DType_T) -> None: ...

class ReadWriteIO(ReadIO[DType_T], WriteIO[DType_T], ABC): ...
```

(Exact class names to be settled in prototype — could equally be
`AttrRIO/AttrWIO/AttrRWIO` to mirror the Attr family.)

- `AttrR(dt, io: ReadIO[DType_T] | None)`, `AttrW(dt, io: WriteIO[...] |
  None)`, `AttrRW(dt, io: ReadWriteIO[...] | None)` — passing a read-only IO
  to an `AttrRW` is a **static** error; abstract methods force subclasses to
  implement the right surface.
- `update_period` lives on `ReadIO` (it describes the IO, not the
  attribute); `ONCE` still supported. `Controller.create_api_and_tasks`
  schedules from `attr.io.update_period` instead of matching ref types.
- `io=None` keeps today's soft behaviour (`AttrRW` self-wires
  setpoint→readback via `_internal_update`; sync-setpoint machinery
  unchanged) — the analogue of `soft_signal_rw`.
- Concrete callback adapters ship for one-offs (the `soft_command`-style
  escape hatch): e.g. `CallbackReadIO(update=cb, update_period=0.2)`.
- Migration is mechanical: an old IO subclass absorbs its ref's fields into
  its constructor and is passed per-attribute.

### 3.4 Target: the temperature controller

Before: 3 cooperating classes (`TemperatureControllerAttributeIORef`,
`TemperatureControllerAttributeIO`, `ios=[...]` registration) + dispatch.
After — one small IO class, everything bound at creation:

```python
class TempIO(ReadWriteIO[float]):
    def __init__(self, conn: IPConnection, param: str, suffix: str = "",
                 update_period: float | None = 0.2):
        super().__init__(update_period=update_period)
        self._conn, self._cmd = conn, f"{param}{suffix}"

    async def update(self, attr: AttrR[float]) -> None:
        resp = await self._conn.send_query(f"{self._cmd}?\r\n")
        await attr.update(attr.dtype(resp.strip("\r\n")))

    async def send(self, attr: AttrW[float], value: float) -> None:
        await self._conn.send_command(f"{self._cmd}={value}\r\n")


class TemperatureRampController(Controller):
    def __init__(self, index: int, conn: IPConnection) -> None:
        super().__init__()
        sfx = f"{index:02d}"
        self.start = AttrRW(Int(), io=TempIO(conn, "S", sfx))
        self.end = AttrRW(Int(), io=TempIO(conn, "E", sfx))
        self.enabled = AttrRW(Enum(OnOff), io=TempIO(conn, "N", sfx))
        self.target = AttrR(Float(precision=3), io=TempIO(conn, "T", sfx))
        self.actual = AttrR(Float(precision=3), io=TempIO(conn, "A", sfx))
        self.voltage = AttrR(Float(precision=3))  # soft: fed by parent @scan
```

An *introspectable* controller instead declares promised children as bare
hints and provisions them in `initialise()`:

```python
class OdinDetector(Controller):
    # declared: must exist after initialise(), typed for users & pyright
    frames: AttrRW[int]
    acquire: Command[[], None]        # typed commands, decision 5

    async def initialise(self) -> None:
        for name, meta in await self._query_parameter_tree():
            self.filler.fill_attribute(name, ...)   # exact API per prototype
        # filler.check_filled() reports any promised-but-missing children
```

### 3.5 `ControllerFiller`

Mirror of `DeviceFiller` (`ophyd_async/core/_device_filler.py` is the
reference implementation — steal its shape):

- scans class hints for `AttrR/W/RW[T]`, `Command[...]`, sub-`Controller` /
  `ControllerVector[T]` types; creates children unfilled; tracks
  filled/unfilled; `check_filled(source)` raises listing what introspection
  failed to provide.
- yields `(child, extras)` so third-party extras vocabularies work
  (decision 3); core consumes none for 1.0.
- subsumes and deletes `HintedAttribute` + `_validate_type_hints`; removes
  the deepcopy half of `_bind_attrs` (method binding for
  `@command`/`@scan` stays).
- trailing-underscore convention adopted from ophyd-async (`stop_` →
  logical name `stop`) for name clashes.

### 3.6 Typed commands

- Lift the zero-arg restriction in `Method`/`Command._validate`; keep the
  captured `inspect.Signature` (already there) as the public
  signature — mirrors `CommandBackend.signature`.
- ControllerAPI exposes the signature; transports declare capability:
  EPICS CA/PVA serve void/void commands and **skip typed ones with a
  warning** (decision 11); Tango/REST/GraphQL serve fully.

---

## 4. What changes in ophyd-async

Deliberately little (it is the reference for the idiom):

- New embedded connector (§5) behind `fastcs-embed` extra.
- `fastcs_connector(uri)` grows scheme dispatch: `pva://` →
  `PviDeviceConnector` (today's behaviour), `tango://` →
  `TangoDeviceConnector`, plus the in-process path taking a `Controller`
  instance directly.
- Follow-up (not blocking): `Device.disconnect()` upstream discussion.

---

## 5. Embedded FastCS: `FastCSDeviceConnector`

```python
from ophyd_async.fastcs import embedded_fastcs_connector

class TempStage(Device):
    ramp_rate: SignalRW[float]
    power: SignalR[float]
    cancel_all: TriggerableCommand
    ramps: DeviceVector[TempRamp]

stage = TempStage(connector=embedded_fastcs_connector(TemperatureController(settings)))
await stage.connect()          # runs controller lifecycle in-process
```

Mechanics:

- `create_children_from_annotations`: `DeviceFiller` with
  `FastCSSignalBackend` / `FastCSCommandBackend` factories, `filled=False`
  (same pattern as `PviDeviceConnector` / `TangoDeviceConnector`).
- `connect_real` (top level): start the `ControllerRunner` —
  `initialise()`, `post_initialise()`, `create_api_and_tasks()`, controller
  `connect()`, initial coros, scan tasks on the running (bluesky) loop —
  then walk the `ControllerAPI` tree and fill children; `check_filled`;
  `set_name`. Idempotent across reconnects. `await connector.shutdown()`
  cancels scan tasks and calls `disconnect()` (decision 8).
- `connect_mock` never touches the controller → mock mode free.

Backend mappings:

| ophyd-async | FastCS |
|---|---|
| `SignalBackend.get_value` | `AttrR.get()` |
| `SignalBackend.set_callback` | `AttrR.add_on_update_callback(cb, always=True)`; stamp `time.time()`, severity 0 (decision 10b) |
| `SignalBackend.put` | `AttrW.put(value)` |
| `SignalBackend.get_setpoint` | new `AttrW` cached setpoint (decision 10a) |
| `SignalBackend.get_datakey` | from `Attribute.datatype` → `SignalMetadata` (units, precision, limits, enum choices) + `make_datakey` |
| `CommandBackend.execute` / `signature` | fastcs `Command.__call__` / `Method` signature |
| `SignalBackend.source` | e.g. `fastcs://<controller-path>.<name>` |
| child Device / `DeviceVector` | sub-`Controller` / `ControllerVector` |
| (not exposed) | `@scan` methods — server-side only |

Datatype mapping: `Int/Float/Bool/String` → `int/float/bool/str`;
`Waveform(array_dtype, shape)` → `Array1D[dtype]`; `Enum(cls)` → the enum
class. **Prototype risk**: ophyd-async constrains enums to `EnumTypes`
(`StrictEnum`/`SubsetEnum`/`SupersetEnum`) while fastcs accepts any
`enum.Enum` — needs a conversion/validation decision. Same for fastcs
`Table` vs ophyd-async `Table` (pydantic-based) — map best-effort, flag
mismatches early.

The stable public Controller interface the connector may use (decision 13):
`ControllerAPI` tree (attributes, command_methods + signatures, sub_apis,
description); `AttrR.get/add_on_update_callback`, `AttrW.put` (+ cached
setpoint), `Attribute.datatype/access_mode/description/group`;
`ControllerRunner` start/stop. Nothing else — no reaching into
`BaseController` internals.

---

## 6. Migration / removal list (FastCS, pre-1.0)

- Remove: `AttributeIORef`, IO registry (`ios=`, `_validate_io`,
  `_connect_attribute_ios`), `AttributeIORefT` TypeVar, `HintedAttribute`,
  class-scope Attribute prototypes + `_bind_attrs` deepcopy.
- Add: R/W/RW `AttributeIO` flavours + `io=` arg; `ControllerFiller` +
  extras mechanism; typed commands; `AttrW` setpoint cache;
  `ControllerRunner`; renamed metadata fields (`precision`, `Limits`-aligned
  names); `Array1D`/`Table` hint spellings.
- Docs: rewrite the "writing a controller" tutorial around the temperature
  controller (§3.4) and an introspectable example; state the class-body /
  instance-scope rule up front, identically to ophyd-async's
  declarative-vs-procedural page.

---

## 7. Worked example pair (the convergence pitch)

Same developer, both projects, same shapes:

```python
# FastCS server                          # ophyd-async client
class Ramp(Controller):                  class RampDevice(Device):
    target: AttrRW[float]                    target: SignalRW[float]
    actual: AttrR[float]                     actual: SignalR[float]
    enable: Command[[], None]                enable: TriggerableCommand
# filled by initialise() introspection   # filled by connector introspection
# procedural: AttrRW(Float(), io=...)    # procedural: soft_signal_rw(float)
# actions: @command / typed args         # actions: Command[P, T] / soft_command
```

### 7.5 The Tango pitch (FastCS vs PyTango)

FastCS must also sell as a way to write Tango Device Servers: more
compelling than PyTango, no more complicated for the simple case.

Where FastCS (with this proposal) beats PyTango:

- **One class, many faces**: Tango + EPICS CA/PVA(+PVI) + REST/GraphQL +
  generated GUI from one controller (see `demo/fastcs.yaml`), plus direct
  embedding in ophyd-async — PyTango gives Tango only.
- **Async-native** throughout; PyTango green modes are second-class (the
  fastcs Tango DSR has to bridge our loop thread-safely already).
- **Composition**: sub-controllers/vectors vs Tango's flat namespace.
- **Introspection-driven attributes** (`ControllerFiller`) — no PyTango
  equivalent.
- **Typing**: dtypes in real annotations, pyright-checked; validation with
  limits/alarms in `DataType`; plain-asyncio testing (no
  `DeviceTestContext`); yaml config with generated JSON schema vs Tango DB
  properties.
- Decisions 5 (typed commands) and 10b (native timestamps) are
  **prerequisites** for this pitch, not nice-to-haves: PyTango has typed
  in/out commands and timestamped attributes already.

Where the un-amended proposal *lost* to PyTango: the trivial case. PyTango
hello-world is one decorated getter (`@attribute def current(self) ->
float: return 2.5`); the harsh split required a method + `CallbackReadIO`
adapter + `__init__` wiring. Decision 14 closes this with property-style
sugar on the existing method-binding machinery:

```python
class PowerSupply(Controller):
    @attr_rw(units="V", update_period=0.5)   # dtype inferred from -> float
    async def voltage(self) -> float:
        return await self._conn.query("V?")

    @voltage.send
    async def voltage(self, value: float) -> None:
        await self._conn.send(f"V={value}")
```

One decorated method for the simple case — dtype stated once, in a real
annotation (better than PyTango's `dtype=` kwargs) — degrading gracefully
into `io=` objects for protocol families and the filler for introspection.
Transport-level Tango gaps noted for later (not proposal blockers):
DevState/Status conventions are currently hardcoded `ON`; event pushing
from `add_on_update_callback` should use native timestamps once 10b lands.
Docs should include a "FastCS for PyTango users" side-by-side page.

---

## 8. Work plan (PR-sized, in order)

**FastCS**
1. `AttributeIO` rework (§3.2–3.4): new R/W/RW IO classes, `io=` arg, delete
   ref machinery, drop second TypeVar, move `update_period`, port demo
   temperature controller + tests. *(Biggest single win; independent of the
   filler.)*
2. Typed commands (§3.6): lift restriction, ControllerAPI signature,
   EPICS skip-with-warning, Tango/REST/GraphQL support.
3. `ControllerFiller` (§3.5): hints-create + filled/unfilled + extras yield;
   delete `HintedAttribute` + class-scope prototypes; port introspecting
   controllers (odin/eiger-style) in dependent repos as validation.
4. `AttrW` setpoint cache; **native timestamps (+severity) on attribute
   updates** (decision 10, 1.0 scope); `ControllerRunner` extraction from
   `FastCS.serve`; document the stable interface (decision 13).
5. Naming pass (decision 12): `precision`, Limits alignment, `Array1D`/
   `Table` hints.
5b. Attr-from-method decorator sugar (decision 14, §7.5): `@attr_r` /
   `@attr_rw` + `.send` on the `Unbound*` bind machinery; sits on PR 1,
   small and independent; include the "FastCS for PyTango users" docs page.

**ophyd-async**
6. `ophyd_async.fastcs` embedded connector (§5) + `fastcs-embed` extra +
   tests against the demo temperature controller (sim serial device exists
   in `fastcs.demo.simulation`). Depends on FastCS 1, 4 (setpoint cache,
   runner); degrade gracefully where possible to prototype earlier.
7. `fastcs_connector` URI scheme dispatch; docs page "embedding FastCS".
8. (Follow-up) `Device.disconnect()` proposal; serve-transports-alongside
   option on the runner (decision 9).

Coordination: 1–4 land in FastCS before 6 is mergeable; prototype 6 against
a FastCS branch to validate the stable interface *before* freezing it.

---

## 9. Key files for the prototyper

**ophyd-async** (`/workspaces/ophyd-async/src/ophyd_async/`)
- `core/_device.py` — `Device`, `DeviceConnector`, `DeviceVector`
- `core/_device_filler.py` — `DeviceFiller` (reference for `ControllerFiller`)
- `core/_signal_backend.py` — `SignalBackend`, `SignalMetadata`, `make_datakey`
- `core/_command.py` — `Command[P, T]`, `CommandBackend`, `TriggerableCommand`, `soft_command`
- `epics/core/_pvi_connector.py` — the unfilled-until-connect workflow
- `tango/core/_base_device.py` — introspection filling + `Annotated` extras consumption
- `fastcs/core.py` — current `fastcs_connector`
- `docs/explanations/declarative-vs-procedural.md`

**FastCS** (`/workspaces/FastCS/src/fastcs/`)
- `attributes/` — `Attribute`, `AttrR/W/RW`, `attribute_io.py`, `attribute_io_ref.py` (to be reworked/removed)
- `controllers/base_controller.py` — `_bind_attrs`, `_find_type_hints`, `_connect_attribute_ios` (bulk of the surgery)
- `controllers/controller.py` — `create_api_and_tasks` (update-period scheduling)
- `methods/` — `Method`/`Command`/`Scan` (typed-command change)
- `datatypes/` — `DataType` family (naming pass)
- `control_system.py` — `FastCS.serve` (extract `ControllerRunner`)
- `demo/controllers.py` — temperature controller (the ergonomics benchmark)

coretl:
This PR is not to be merged, but is there so we can make drive-by comments on lines of code as we see them be accepted in PRs into the refactor branch. When we are happy we will flip this PR to non-draft and this will become what is merged.


docs/explanations/decisions/0013-declarative-procedural-split-and-controller-filler.md
```diff
@@ -0,0 +1,158 @@
...
+Example, before and after:
+
+```python
+# Before: class-scope instance, deepcopy'd per-instance
+class TemperatureRampController(Controller):
+    start = AttrRW(Int(), io_ref=TemperatureControllerAttributeIORef(name="S"))
+
+# After: bare hint, filled procedurally
+class TemperatureRampController(Controller):
+    start: AttrRW[int]
```
coretl: This is not a good example of bare hint. This should either be an Eiger-like example where we have type hints, and a filler is called in init to ensure the live device has those attributes, or this is a TemperatureRampController where there is no type hint at all, only an `__init__` assignment

docs/explanations/decisions/0013-declarative-procedural-split-and-controller-filler.md
```diff
@@ -0,0 +1,158 @@
...
+
+Introspecting controllers (`fastcs-eiger`, `fastcs-secop`,
+`fastcs-PandABlocks`, `fastcs-catio`'s dynamic path) keep working exactly as
+today's `initialise()` + `add_attribute` pattern, but the fully-dynamic case
+where the *set* of attributes is not known until a network round-trip
+completes (`fastcs-PandABlocks`, `fastcs-secop`) needs `ControllerFiller` to
+support filling children that were never hinted at all — mirroring
+`DeviceFiller.fill_child_signal`'s "no annotation existed, introspection
+added an undeclared attribute" path. This is a harder requirement than most
+of ophyd-async's own connectors exercise (PVI and Tango both fill *some*
```
coretl: This is not harder than ophyd-async, it is already supported there. `Device(connector=PviConnector(prefix))` will do this for instance, no type hint required.

docs/explanations/decisions/0013-declarative-procedural-split-and-controller-filler.md
```diff
@@ -0,0 +1,158 @@
...
+  it interacts with the stable `ControllerAPI` surface consumed by the
+  embedded ophyd-async connector.
+
+## Open questions
+
+1. Does `ControllerFiller` need to support "no hints exist at all — build the
+   entire attribute tree from introspected data" (the `fastcs-PandABlocks`
+   and `fastcs-secop` case), or is some minimal static shape (even just a
+   marker on the `Controller` subclass) always required? `DeviceFiller` has
+   no precedent for the fully-hint-free case.
```
coretl: Yes, this should support "no hints at all"

docs/explanations/decisions/0013-declarative-procedural-split-and-controller-filler.md
```diff
@@ -0,0 +1,158 @@
...
+2. `fastcs-catio`'s dynamic path builds whole controller *classes* at runtime
+   via `type(...)` from YAML definitions, before any instance (and hence any
+   `ControllerFiller`) exists. Is that pattern still supported, unsupported,
+   or does it need to move to instance-level dynamic attribute construction
+   under the new model?
```
coretl: Unsupported. A bare `Controller` should allow attributes to be added on it from the outside, which is what the Fillers will do.

docs/explanations/decisions/0013-declarative-procedural-split-and-controller-filler.md
```diff
@@ -0,0 +1,158 @@
...
+3. `fastcs-eiger`'s `OdinController.initialise()` constructs new attributes
+   that reference sibling sub-controllers' attributes, assuming those
+   sub-controllers already exist. Does `ControllerFiller` impose an
+   ordering/dependency mechanism between sibling children, or is this left
+   as an `initialise()` implementation detail (call `super().initialise()`
+   first)?
```
coretl: The rule is: at the end of `__init__` any Attribute you reference in code (and so has a type hint) *must* exist. This is the rule in ophyd-async for Signals too. This means that initialise can happen in parallel, only `__init__` is serial.

docs/explanations/decisions/0013-declarative-procedural-split-and-controller-filler.md
```diff
@@ -0,0 +1,158 @@
...
+4. Should `check_filled` be able to distinguish "this hinted child is
+   optional" (ophyd-async's `Optional[X]` convention), or does FastCS treat
+   every hint as required for 1.0?
```
coretl: Add support for `Optional`

docs/explanations/decisions/0013-declarative-procedural-split-and-controller-filler.md
```diff
@@ -0,0 +1,158 @@
...
+5. Exact `ControllerFiller` method names/signatures are left to the
+   prototype — should they mirror `DeviceFiller`'s names 1:1
+   (`fill_child_signal` → `fill_child_attribute`?) for discoverability by
+   developers who know both libraries, or diverge where FastCS's vocabulary
+   (`Attribute` vs `Signal`) differs?
```
coretl: Follow the structure, the names are less important. Architectural similarity is important, names should only match if they fit.

docs/explanations/decisions/0014-attribute-io-rw-rework.md
```diff
@@ -0,0 +1,186 @@
...
+        await attr.update(float(resp))
+
+ramp_rate = AttrRW(Float(), io_ref=TempIORef(name="R"))
+# ... elsewhere: Controller(ios=[TempIO(conn)])
+
+# After
+class TempIO(ReadWriteIO[float]):
+    def __init__(self, conn: IPConnection, name: str, update_period=0.2):
+        super().__init__(update_period=update_period)
+        self._conn, self._name = conn, name
```
coretl: Make this a dataclass

docs/explanations/decisions/0014-attribute-io-rw-rework.md
```diff
@@ -0,0 +1,186 @@
...
+
+```python
+# Before (ADR 9 shape)
+class TempIORef(AttributeIORef):
+    name: str
+
+class TempIO(AttributeIO[float, TempIORef]):
+    async def update(self, attr: AttrR[float, TempIORef]) -> None:
+        resp = await self._conn.send_query(f"{attr.io_ref.name}?\r\n")
+        await attr.update(float(resp))
```
coretl, line 247: Where's the send method?

docs/explanations/decisions/0014-attribute-io-rw-rework.md
```diff
@@ -0,0 +1,186 @@
...
+  runtime check as well (e.g. in `post_initialise`) is an open question.
+- [ADR 12](0012-attribute-io-naming-convention.md)'s guidance (subclass to
+  get a shorter driver-local name) still applies to the new `ReadIO`/
+  `WriteIO`/`ReadWriteIO` names.
+
+## Open questions
+
+1. Final class names: `ReadIO`/`WriteIO`/`ReadWriteIO` vs. `AttrRIO`/
+   `AttrWIO`/`AttrRWIO` (mirroring the `Attr` family) vs. something else
+   entirely — see [ADR 17](0017-naming-pass.md).
```
coretl: @shihab-dls what do you reckon?

docs/explanations/decisions/0014-attribute-io-rw-rework.md
```diff
@@ -0,0 +1,186 @@
...
+2. What is the public replacement for `fastcs-secop`'s
+   `_call_sync_setpoint_callbacks` workaround? Does `WriteIO.send` get an
+   optional `sync_setpoint` callback argument, or does `AttrW.put` grow a
+   public method IO authors can call from `send`?
```
coretl: @shihab-dls I can't answer this. @Tom-Willemsen your input would be good here too.

docs/explanations/decisions/0014-attribute-io-rw-rework.md
```diff
@@ -0,0 +1,186 @@
...
+3. Should there be a runtime check (e.g. at `post_initialise`) that
+   catches "read-only IO passed to a write-capable `Attr`" for cases the
+   static type checker cannot see (e.g. an `Any`-typed IO built
+   dynamically, as in `fastcs-secop`'s and `fastcs-PandABlocks`'s
+   introspection-driven construction)? Both of those drivers build
+   attributes and their IO from runtime data where static checking cannot
+   help.
```
coretl: Yes

docs/explanations/decisions/0014-attribute-io-rw-rework.md
```diff
@@ -0,0 +1,186 @@
...
+4. `fastcs-PandABlocks`'s `UnitsIO.send` mutates a sibling attribute's
+   datatype and `fastcs-catio` recovers per-attribute metadata via
+   `attribute.io_ref` from *outside* the attribute's own `send`/`update`
+   (`panda_controller.py:_coerce_value_to_panda_type`). With `io_ref`
+   removed, what is the sanctioned way to recover an attribute's IO-specific
+   metadata (e.g. `attr.io` becoming a public, typed property)?
```
coretl: `attr.io` becomes public

docs/explanations/decisions/0014-attribute-io-rw-rework.md
```diff
@@ -0,0 +1,186 @@
...
+5. Do we ship `CallbackReadIO`/`CallbackWriteIO` in `fastcs` core for 1.0, or
+   leave the "no subclass needed" one-off case entirely to driver authors
+   using `io=None` plus manual `set_update_callback`?
```
coretl: Combine `CallbackReadIO` and `CallbackWriteIO` use cases into the pyTango style decorator. Maybe the same decorator can be used for both? Then these are equivalent:
```python
@attr
def temperature(self) -> float:
    return lib.get_temperature()

# or

self.temperature = attr(getter=lib.get_temperature)
```

docs/explanations/decisions/0015-typed-commands.md
```diff
@@ -0,0 +1,102 @@
...
+  as [ADR 13](0013-declarative-procedural-split-and-controller-filler.md).
+- Command args/return values need datatype validation analogous to
+  `Attribute`'s `DataType.validate` — whether they reuse the `DataType`
+  family directly or a separate mechanism is an open question.
+
+## Open questions
+
+1. Do command arguments/return values validate through the same `DataType`
+   family attributes use, or is a separate (lighter-weight, since there's no
+   "current value" to cache) validation path introduced?
```
coretl: I would treat this the same as Attr: if you have no metadata, use the python type. I'm tempted to say that we drop the FastCS DataType completely, and instead go with some `*Meta` typed dicts, which can be unpacked into the constructor via typing.overrides:
```python
temperature: A[AttrR[float], FloatMeta(precision=3, units="deg")]
# or
self.temperature = AttrR(float, precision=3, units="deg", io=...)
```
Which can then be supplied as kw-only to the `attr` factory callable too:
```python
@attr(precision=3, units="deg")
def temperature(self) -> float:
    return lib.get_temperature()

# or

self.temperature = attr(getter=lib.get_temperature, precision=3, units="deg")
```

docs/explanations/decisions/0015-typed-commands.md
```diff
@@ -0,0 +1,102 @@
...
+2. For `fastcs-secop`-style dynamically-typed commands, what's the
+   recommended pattern — `Command[Any, Any]` with manual validation inside
+   the handler, or a documented way to construct a `Command[P, T]` with `P`/
+   `T` determined at runtime (which conflicts with normal generic typing)?
```
coretl: I think we allow type hints of varying compexity:
```python
fully_known: Command[[float], None]  # I know it takes a float and returns nothing, validate this
partially_known: Command[Any, Any]  # I know it must be a command, but introspect it to find out what it takes 
```
I would also like to support keyword args here, but the language is a little limited. We might need to play some funny tricks with `if TYPE_CHECKING` and a stub so we could do something like:
```python
def fn_spec(named_arg: float) -> int: ...

if TYPE_CHECKING:
    def command_spec(spec: Callable[P, T]) -> Command[P, T]: ...
    # and something at runtime, but not sure what

kw_known = command_spec(fn_spec)
```

See https://github.com/bluesky/ophyd-async/pull/1138#discussion_r2965959411

@Tom-Willemsen are there really secop devices where you know that it has a Command, but you don't know its type until runtime?

docs/explanations/decisions/0015-typed-commands.md
```diff
@@ -0,0 +1,102 @@
...
+3. Exactly what should the EPICS skip-with-warning message say, and where —
+   at controller construction, at `post_initialise`, or lazily the first
+   time a typed command is looked up by the transport?
```
coretl: It should do this on IOC startup. This is post controller construction, when the fully populated controllers are passed to the transports to serve

docs/explanations/decisions/0015-typed-commands.md
```diff
@@ -0,0 +1,102 @@
...
+4. Should typed commands support partial typing (e.g. typed arguments but
+   void return, or vice versa), or is it all-or-nothing relative to
+   `Command[[], None]`?
```
coretl: Args and Returns should be treated separately.

Args can be `[]` (nothing), `[DT1, DT2, ...]` (positional or kw args all of known types), or `Any` (not known which of the former it is until runtime)

Returns can be `None` (nothing), `DT` (a single return value) or `Any` (not known which of the former it is until runtime)

docs/explanations/decisions/0015-typed-commands.md
```diff
@@ -0,0 +1,102 @@
...
+5. Does the REST/GraphQL/Tango serialisation of complex argument/return
+   types (numpy arrays, `Enum`, `Table`) reuse existing `DataType`
+   serialisation code from attributes, and if so does that argue for
+   sharing more machinery between `Attribute` and `Command` than they do
+   today?
```
coretl: Yes

docs/explanations/decisions/0016-setpoint-cache-timestamps-and-controller-runner.md
```diff
@@ -0,0 +1,112 @@
...
+  transport-specific follow-up work, not part of this ADR.
+- The embedded ophyd-async connector becomes buildable against a documented,
+  narrow surface instead of `BaseController` internals — see
+  [ADR 19](0019-embedded-ophyd-async-connector.md).
+
+## Open questions
+
+1. Setpoint cache accessor name and shape — `AttrW.setpoint` property,
+   `AttrW.get_setpoint()` method (mirroring `SignalBackend.get_setpoint()`),
+   or folded into `AttrW.put`'s return value?
```
coretl: `.get_setpoint()` method

docs/explanations/decisions/0016-setpoint-cache-timestamps-and-controller-runner.md
```diff
@@ -0,0 +1,112 @@
...
+2. Timestamp/severity type — reuse a existing convention (e.g.
+   `ophyd_async`/bluesky's `Reading`/event-model shape) or define a
+   FastCS-native pair? Decision 12 of #388 already aligns numeric limits
+   naming with event-model `Limits` — should timestamps/severity follow the
+   same alignment for consistency?
```
coretl: Follow `Reading` shape, but don't share any code

docs/explanations/decisions/0016-setpoint-cache-timestamps-and-controller-runner.md
```diff
@@ -0,0 +1,112 @@
...
+3. Severity: what are the FastCS-native severity levels, and do they map
+   1:1 to EPICS alarm severities, or is EPICS's severity model transport-
+   specific with FastCS defining its own smaller/different vocabulary?
```
coretl: Make a fastcs enum, but use the same strings as EPICS

docs/explanations/decisions/0016-setpoint-cache-timestamps-and-controller-runner.md
```diff
@@ -0,0 +1,112 @@
...
+4. Exact `ControllerRunner` API shape — a class with `start()`/`stop()`, or
+   `async` context-manager semantics (`async with runner:`)? The embedded
+   connector needs idempotent start across reconnects; does the chosen shape
+   make idempotency the caller's responsibility or the runner's?
```
coretl: Class with `start` and `stop`. Context manager only if it makes sense for the `FastCS()` case as ophyd-async will use `start` and `stop`. Idempotency is caller's responsibility.

docs/explanations/decisions/0016-setpoint-cache-timestamps-and-controller-runner.md
```diff
@@ -0,0 +1,112 @@
...
+5. Does `ControllerRunner` own reconnect logic (calling `Controller.reconnect()`
+   on scan-task failure, as `Controller._create_periodic_scan_coro` does
+   today), or does that stay controller-specific and out of the runner's
+   documented surface?
```
coretl: `ControllerRunner` owns the whole life-cycle, including reconnect

docs/explanations/decisions/0017-naming-pass.md
```diff
@@ -0,0 +1,105 @@
...
+  attribute declarations; procedural construction with `Waveform(...)`/
+  `Table(...)` DataType instances is unchanged.
+
+## Open questions
+
+1. Does the Limits alignment keep four flat fields (just renamed to match
+   event-model terms) or restructure into an actual `Limits`-like nested
+   object? The latter is a bigger, more disruptive change to
+   `DataType.validate` and every downstream driver constructing `Float(...)`
+   with keyword limits.
```
coretl: Nested

docs/explanations/decisions/0017-naming-pass.md
```diff
@@ -0,0 +1,105 @@
...
+2. Which event-model `Limits` categories does FastCS need —
+   control/display/alarm/warning all four, or a subset? EPICS records only
+   naturally distinguish alarm vs. display/control limits; does the mapping
+   from four FastCS fields to N event-model categories lose or need to
+   invent information for some transports?
```
coretl: All needed but optional. Some rules:
- Don't supply any, they are all unbounded
- Supply `Display` but not `Control`, then `Control` inherits it for writeable
- Supply `Alarm` but not `Warning`, then `Warning` inherits it
- Supply `Alarm` and `Warning`, then assert `Warning` is inside or equal to `Alarm`
- Otherwise the unspecified limits are unbounded

docs/explanations/decisions/0017-naming-pass.md
```diff
@@ -0,0 +1,105 @@
...
+3. Is `precision` an `int` (decimal places, as `prec` is today) or does
+   aligning with event-model conventions change its meaning/type too?
```
coretl: Keep an int

docs/explanations/decisions/0017-naming-pass.md
```diff
@@ -0,0 +1,105 @@
...
+4. For `Array1D`/`Table` hints: is `Array1D[np.int32]` a real usable type at
+   both class-definition time (for `ControllerFiller` to scan) and at
+   type-checking time (for pyright), or a `TypeAlias`/`Annotated` wrapper
+   around `Waveform`? What does the two-way mapping (hint → `ControllerFiller`
+   constructs a `Waveform`; introspection-provisioned `Waveform` → does the
+   hint still validate it, per decision in
+   [ADR 13](0013-declarative-procedural-split-and-controller-filler.md)
+   open question 1) look like precisely?
```
coretl: If we ditch `DataType` then this all falls out in the wash. `Array1D` is the hint, and the runtime structure that is passed around as datatype

docs/explanations/decisions/0017-naming-pass.md
```diff
@@ -0,0 +1,105 @@
...
+5. Should this rename land in the same PR as
+   [ADR 14](0014-attribute-io-rw-rework.md) (since both touch `DataType`-
+   adjacent code and every downstream driver already has to touch these
+   files), or stay a separate, later PR per §8 work-plan ordering (item 5,
+   after items 1-4)?
```
coretl: Whatever works best at implementation time.

docs/explanations/decisions/0018-attr-decorator-sugar.md
```diff
@@ -0,0 +1,131 @@
...
+- Per #388 §8 item 5b, this sits on PR 1 (the `AttributeIO` rework) —
+  small and independent of the `ControllerFiller` work — so it can land
+  early and not block on [ADR 13](0013-declarative-procedural-split-and-controller-filler.md).
+
+## Open questions
+
+1. Exact decorator names — `@attr_r`/`@attr_rw` as #388 proposes, or
+   something more explicit (`@readable_attribute`?) — and whether a
+   write-only `@attr_w` variant is worth adding for symmetry given `AttrW`
+   without a paired getter is a rarer shape in practice.
```
coretl: Tempted by mirroring @property a bit closer:
```python
class PowerSupply(Controller):
    @attr(units="V", update_period=0.5)   # dtype inferred from -> float, deferred creation until init
    async def voltage(self) -> float:
        return await self._conn.query("V?")

    @voltage.setter
    async def voltage(self, value: float) -> None:  # now deferred creation with make an AttrRW
        await self._conn.send(f"V={value}")
```

As `AttrW` is much rarer then people have to write it longhand

docs/explanations/decisions/0018-attr-decorator-sugar.md
```diff
@@ -0,0 +1,131 @@
...
+2. How are `min`/`max`/`precision` (post [ADR 17](0017-naming-pass.md))
+   and other `DataType`-level metadata passed through the decorator's
+   keyword arguments — do they get their own decorator kwargs, or does the
+   decorator only take IO-shaped kwargs (`update_period`) and require
+   dropping to explicit `AttrRW(...)` construction for richer datatype
+   metadata?
```
coretl: decorator kwargs

docs/explanations/decisions/0018-attr-decorator-sugar.md
```diff
@@ -0,0 +1,131 @@
...
+3. Does `@attr_rw` support the `Array1D`/`Table` hint spellings from
+   [ADR 17](0017-naming-pass.md), or is decorator sugar scoped to scalar
+   datatypes only for 1.0?
```
coretl: Support ADR17 hints

docs/explanations/decisions/0018-attr-decorator-sugar.md
```diff
@@ -0,0 +1,131 @@
...
+4. Should `ControllerFiller` treat `@attr_rw`-decorated methods specially
+   (they don't need filling — they're already fully constructed at bind
+   time), or are they simply invisible to the filler the same way
+   `@command`/`@scan` are today?
```
coretl: Treat them specially, they are defined already so don't shadow them, and raise if there is a clash between introspected name and decorated name

docs/explanations/decisions/0018-attr-decorator-sugar.md
```diff
@@ -0,0 +1,131 @@
...
+5. Does the getter's docstring become the attribute's `description`,
+   mirroring how `Method._docstring` already captures `getdoc(fn)` for
+   `@command`/`@scan`?
```
coretl: Yes

docs/explanations/decisions/0019-embedded-ophyd-async-connector.md
```diff
@@ -0,0 +1,168 @@
...
+  (`fastcs.demo.simulation`) is the existing sim device ophyd-async tests
+  against — no new simulated device is needed for the first cut.
+
+## Open questions
+
+1. Enum conversion: does the connector require FastCS `Enum` datatypes used
+   with embedding to be `StrictEnum`/`SubsetEnum`/`SupersetEnum` subclasses
+   (pushing a constraint back onto FastCS driver authors who want embedding
+   support), or does it do runtime conversion/wrapping, and what happens to
+   values that don't fit ophyd-async's stricter model?
```
coretl: Un-type-hinted classes will be fine: introspect the Enum at runtime, and drop to string datatype retaining choices as metadata. It's the type hinted classes that need thinking about, what do we type hint with? To start with let's just require people to duplicate, make a `StrictEnum` (or `Subset` or `Superset`) just like they would if FastCS was remote, then revisit when we have some use cases.

docs/explanations/decisions/0019-embedded-ophyd-async-connector.md
```diff
@@ -0,0 +1,168 @@
...
+2. `Table` mapping: is a real bidirectional pydantic-model ↔ fastcs-`Table`
+   converter in scope for the first cut, or is `Table` explicitly
+   unsupported/best-effort-only initially, with a hard error on mismatch
+   rather than silent coercion?
```
coretl: Converter is in scope for first cut. Use this as an opportunity to bring FastCS and ophyd-async Table implementations closer together

docs/explanations/decisions/0019-embedded-ophyd-async-connector.md
```diff
@@ -0,0 +1,168 @@
...
+3. Where does `@scan`-derived state that isn't exposed as a `Signal` go —
+   is it simply invisible to ophyd-async (server-side only, as the mapping
+   table states), or does some `@scan` output need a path to surface as a
+   `SignalR` (e.g. `fastcs-eiger`'s `update_voltages` `@scan` feeding
+   per-ramp `AttrR`s — those `AttrR`s are visible, but would a *pure*
+   `@scan`-only value ever need exposing)?
```
coretl: I don't understand this. Isn't `@scan` just driving the update? Doesn't all the data have to go in `Attr` instances anyway which are mapped to `Signal`s?

@shihab-dls have I missed something?

docs/explanations/decisions/0019-embedded-ophyd-async-connector.md
```diff
@@ -0,0 +1,168 @@
...
+4. How does `embedded_fastcs_connector` handle a `Controller` that raises
+   during `initialise()` (e.g. a device that's unreachable at embed time)
+   — does `connect_real` propagate the exception directly to
+   `Device.connect()`, retry, or something else?
```
coretl: FastCS needs a `ConnectionFailedError` that it can raise if the device didn't respond. This will be converted into a `NotConnectedError` by the connector. It will then continue retrying to connect in the background. All other errors should be surfaced without conversion.

docs/explanations/decisions/0019-embedded-ophyd-async-connector.md
```diff
@@ -0,0 +1,168 @@
...
+5. Should the embedded connector's shutdown (`atexit` + explicit
+   `await connector.shutdown()`) also be triggered by ophyd-async's own
+   `Device.disconnect()` once that upstream work lands (#388 §8 item 8), and
+   does that imply `ControllerRunner.stop()` needs to be safely callable
+   from a synchronous `atexit` context as well as an async one?
```
coretl: Ditch the `Device.disconnect` proposal. `Device.connect(force_reconnect=True)` acts as a "reconnect". The only time we want to disconnect is atexit, so leave it there

docs/explanations/decisions/0014-attribute-io-rw-rework.md
Tom-Willemsen: I recall we needed this because after a `caput`, doing a `caget` on the setpoint PV was not reflecting the just-sent setpoint, which is the behaviour we'd expect from a conventional IOC and what we've trained our users to expect in GUI etc. Think this would have been using CA transport, though possible we were seeing it on PVA instead...

My question back to you is, is that considered a bug in the transport layer, or intentional behaviour? Are there use-cases for the setpoint PV _not_ showing the most-recently-written value?
- If it's intentional then I feel we need _some_ public API to be able to set the setpoint value to reflect the just-sent value.
- If it's a bug in the transport, then once that is fixed we wouldn't need this API downstream.

docs/explanations/decisions/0015-typed-commands.md
Tom-Willemsen: SECoP self-describes its commands, including argument and return types, over-the-wire in response to a `describe` message. A fragment of the `datainfo` (the datatype for a single command-type parameter, sent over-the-wire) looks like:

```
{"type": "command", "argument": {"type": "bool"}, "result": {"type": "int"}}

--- or ---

{"type": "command",
 "argument": {"type": "struct", "members": {"p": {"type": "double"},
                                            "i": {"type": "double"},
                                            "d": {"type": "double"}}},
 "result": {"type": "tuple", "members": [{"type": "int"}, {"type": "string"}]}}
```

(from [secop docs](https://sampleenvironment.github.io/secop-site/specification/datainfo.html#command-command))

We don't "know that it has a command" statically up-front - we only know that at runtime, once we've sent the device a `describe` message and it has replied with a description of all of its attributes and commands. In fact with SECoP we know _nothing_ up-front, there are no static attributes or commands that we would define with a type-hint. It is entirely configured from the response to an over-the-wire `describe` message.

The current approach to commands in `fastcs-secop` is to add a [`SecopCommandController`](https://github.com/ISISComputingGroup/fastcs-secop/blob/main/src/fastcs_secop/_controllers.py#L43) sub-controller whenever we see a `command`-typed object in the output of `describe`. This adds PVs for argument (if non-void), return value (if non-void), and execute (unconditionally).

docs/explanations/decisions/0017-naming-pass.md
Tom-Willemsen: I might not be following correctly, but just to note SECoP carries information about e.g. max length of an array in its `datainfo` (so sent over-the-wire, only available at runtime).

We'd still need the capability to get the max length of an array (which eventually makes it to `NORD` of an EPICS CA waveform) at runtime somehow.

Not 100% clear to me how this interacts with the proposal to move `Waveform(..., shape=...)` to `Array1D[np.int32]`?

docs/explanations/decisions/0019-embedded-ophyd-async-connector.md
Tom-Willemsen: Again not sure I'm following the proposed changes properly, but just to note `fastcs-secop` [_manufactures_ enums at runtime](https://github.com/ISISComputingGroup/fastcs-secop/blob/main/src/fastcs_secop/_util.py#L158) (as the choices come from parsing SECoP's `datainfo` over-the-wire). So need to ensure that introspection still copes with this sort of case?

docs/explanations/decisions/0015-typed-commands.md
Tom-Willemsen: What I'll also say though, is we need these commands to work with CA and PVA (to the extent that CA/PVA support the argument and return types). That might mean that we stick with our existing `SecopCommandController` approach/workaround which splits arguments/return value/execute into 3 PVs rather than a generic-typed command, if the CA and PVA transports don't support those generic-typed commands.

coretl: We can still make the CA and PVA transport explode commands with args into multiple PVs though, we're talking about what the most natural structures are in FastCS

docs/explanations/decisions/0014-attribute-io-rw-rework.md
coretl: @shihab-dls this sounds like the bug you've been fixing, the sequence should now be:
- User does caput/pvput setpoint
- Transport posts update on setpoint
- Controller calls Attr send method
- If it fails then setpoint goes into alarm

This means that as soon as the setpoint is put in the UI, it is reflected in the setpoint PV, which might later turn red if the controller rejects it

@shihab-dls is that right?

docs/explanations/decisions/0019-embedded-ophyd-async-connector.md
shihab-dls: No necessarily; a `@scan` decorated method will just result in a coroutine of that method being run periodically. An `AttrR` with an update method on it's referenced IO will result in a scan coroutine of the update callback being run periodically, but will also be bound to the Attr (and therefore exposed as a Signal in ophyd-async). Both types of coroutines are collected and run in `create_api_and_tasks()` in `controller.py`, but `@scan` is purely internal. Good to note that a `@command` decorated method is different, as this **does** result in an `AttrW` being created and exposed.

docs/explanations/decisions/0014-attribute-io-rw-rework.md
shihab-dls: For PVA, the order of operations mentioned above is correct. For CA, however, the post on the PV is only done after the update callback is complete (which is where our alarms are set), so you'll see:

- User does caput setpoint
- Transport runs update callbacks, where Attr send method is called
- if it fails then setpoint goes into alarm
- Transport posts update on setpoint

on CA then, If you have a long running send, you won't see the setpoint change until after that send is complete.

docs/explanations/decisions/0019-embedded-ophyd-async-connector.md
coretl: Thanks — folded into ADR 0019 in 54522d75: \`@scan\` recorded as a purely-internal periodic coroutine that surfaces no Signal, with the \`@command\`→\`AttrW\`→exposed-Signal contrast called out explicitly. This closes the last open question on 0019.

docs/explanations/decisions/0014-attribute-io-rw-rework.md
coretl: Thanks for the CA/PVA detail — captured in ADR 0014 (54522d75) under *Resolved in review*: \`set()\` caching \`.setpoint\` is an attribute-cache guarantee only; PVA posts the setpoint immediately (may later alarm) while CA posts it only after the update callback completes, so a long-running setter delays the CA-visible setpoint. Realigning CA to PVA's post-before-send ordering is noted as a transport-layer follow-up, not gating this rework.

Issue #392, opened 2026-07-20
getter/setter IO rework; remove AttributeIORef and AttributeIO


coretl:
**Foundational** — unblocks most of the refactor.

> **Rescoped 2026-08-03.** This issue was too large to land green in one session, so per its own "Note on size" it has been split. The `DataType` removal / `*Meta` typed dicts half — and with it the naming pass folded in from #396 — is now **#413**. What remains below is the getter/setter half, delivered by #412.

## Scope

- **Per-attribute IO as `getter`/`setter` callables** on the constructors: `AttrR(getter=…)`, `AttrW(setter=…)`, `AttrRW(getter=…, setter=…)`. Access mode is enforced by *which parameters exist* — there is **no** `io=` object and **no** `ReadIO`/`WriteIO`/`ReadWriteIO` hierarchy (superseded, ADR 0014 review update 2026-07-22).
  - `getter() -> T | Update[T]`; `setter(value) -> None | T | Update[T]` (a returned value is the device's *accepted*/clamped readback → updates readback + `AttrW` setpoint cache; the sanctioned replacement for `fastcs-secop`'s private `_call_sync_setpoint_callbacks`).
  - `Update[T]` = `readback: T`, `timestamp: float | None` (`None` ⇒ framework stamps receive-time), `setpoint: T | None` (`None` ⇒ leave the cached setpoint alone). `severity` is #395.
  - **Datatype optional when a getter/setter is given** — inferred from the getter's return annotation (or setter param), unwrapping `Update[T]` → `T`. Not inferable (`-> Any`, unannotated) ⇒ positional datatype required (fail-fast at construction). The per-datatype `Unpack[*Meta]` static check that keys off the inferred type is #413.
  - **Polling schedule travels with the getter**: a bare getter is read `ONCE`; `Polled(getter, period=0.2)` polls; `Polled(getter, period=None)` is on-demand only. There is no separate `poll_period` constructor argument. **No getter** ⇒ soft, value pushed via `@scan`/`attr.update()`.
  - Soft = absence of getter/setter (`AttrRW(float)` self-wires setpoint→readback). No `io=None` sentinel.
[...]

PR #409, opened 2026-07-22
demo: use ControllerVector for temperature ramp sub-controllers

### LINKED ISSUE #390: Example — composition & methods (temperature controller)
Part of the tutorial examples (#388). The **composition & methods** example — sub-controllers / `ControllerVector`, `@scan`, `@command` — on the temperature backend.

(Formerly "the `io=` pattern"; `io=` objects are gone, replaced by getter/setter callables, so there is nothing to "factor into". This module's role is now **composition**, and it is walked as part of the declarative tutorial — see #405 / #408. It is not its own tutorial.)

## Scope
- **Evolve `src/fastcs/demo/controllers.py` in place** (the demo temperature controller): multiple ramp **sub-controllers** in a `ControllerVector`, a voltage-distributing `@scan`, and a cancel-all `@command`. IO is `getter`/`setter` callables.
- **Baseline now (deliberately messy):** current API (`Float(...)`, `io_ref=...`).
- **Clean form (when #392 lands):** `AttrRW(float, precision=3, getter=…, setter=…)` — migrated by #392.
- Unit tests against the sim.

## Acceptance
`uv run --locked tox -e pre-commit,type-checking,tests` green.

ADR: 0014
Parent: #388
Blocked by: none

coretl:
Closes #390

Evolves `src/fastcs/demo/controllers.py`'s composition example onto the documented `ControllerVector` pattern (see `docs/explanations/controllers.md`) instead of a manual `list` + `add_sub_controller` loop, and adds unit tests exercising `cancel_all` and the voltage-distributing `@scan` against a mocked `IPConnection`.

This is the current-API baseline described in the issue (`Float(...)`, `io_ref=...`); the getter/setter cleanup lands separately when #392 lands.

1. `uv run pytest tests/demo/test_controllers.py -v`
2. Confirm the temperature controller demo still runs against the sim (`python -m fastcs.demo run src/fastcs/demo/fastcs.yaml`)

### Checks for reviewer

### Notes
- Verified locally with `uv run --locked tox -e pre-commit,type-checking,tests`. `pre-commit` and `type-checking` pass. In the sandbox this ran in, the `tests` env's plain `pytest` collection also picks up `docs/conf.py` (needs outbound network) and `tests/benchmarking` (needs a PVA-capable socket family), both of which fail to even collect for reasons unrelated to this change. Excluding those two paths, `pytest src tests --ignore=tests/benchmarking` passes 310/320 with only PVA (`p4p`) IOC tests failing on `RuntimeError: Address family not supported by protocol` — the same known sandbox limitation. Real CI covers these.

---

### DIFF
--- a/src/fastcs/demo/controllers.py
+++ b/src/fastcs/demo/controllers.py
@@ -8,7 +8,7 @@ import numpy as np
 
 from fastcs.attributes import AttributeIO, AttributeIORef, AttrR, AttrRW, AttrW
 from fastcs.connections import IPConnection, IPConnectionSettings
-from fastcs.controllers import Controller
+from fastcs.controllers import Controller, ControllerVector
 from fastcs.datatypes import Enum, Float, Int, Waveform
 from fastcs.logging import logger
 from fastcs.methods import command, scan
@@ -80,15 +80,16 @@ class TemperatureController(Controller):
 
         self._settings = settings
 
-        self._ramp_controllers: list[TemperatureRampController] = []
-        for index in range(1, settings.num_ramp_controllers + 1):
-            controller = TemperatureRampController(index, self.connection)
-            self._ramp_controllers.append(controller)
-            self.add_sub_controller(f"R{index}", controller)
+        self.ramps: ControllerVector[TemperatureRampController] = ControllerVector(
+            {
+                index: TemperatureRampController(index, self.connection)
+                for index in range(1, settings.num_ramp_controllers + 1)
+            }
+        )
 
     @command()
     async def cancel_all(self) -> None:
-        for rc in self._ramp_controllers:
+        for rc in self.ramps.values():
             await rc.enabled.put(OnOffEnum.Off, sync_setpoint=True)
             # TODO: The requests all get concatenated and the sim doesn't handle it
             await asyncio.sleep(0.1)
@@ -118,14 +119,14 @@ class TemperatureController(Controller):
 
         await self.voltages.update(voltages)
 
-        for index, controller in enumerate(self._ramp_controllers):
+        for index, controller in self.ramps.items():
             self.log_event(
                 "Update voltages",
                 topic=controller.voltage,
                 query=query,
                 response=voltages,
             )
-            await controller.voltage.update(float(voltages[index]))
+            await controller.voltage.update(float(voltages[index - 1]))
 
 
 class TemperatureRampController(Controller):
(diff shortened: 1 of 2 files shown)


src/fastcs/demo/controllers.py
```diff
@@ -80,15 +80,16 @@ def __init__(self, settings: TemperatureControllerSettings) -> None:
 
         self._settings = settings
 
-        self._ramp_controllers: list[TemperatureRampController] = []
-        for index in range(1, settings.num_ramp_controllers + 1):
-            controller = TemperatureRampController(index, self.connection)
-            self._ramp_controllers.append(controller)
-            self.add_sub_controller(f"R{index}", controller)
+        self.ramps: ControllerVector[TemperatureRampController] = ControllerVector(
```
coretl: Do we need the type hint here?

coretl: Good catch, removed - pyright infers it fine from the dict literal.

---

tests/demo/test_controllers.py
```diff
@@ -0,0 +1,59 @@
...
+async def test_cancel_all_disables_every_ramp(controller: TemperatureController):
+    controller.connection.send_command = AsyncMock()  # type: ignore[method-assign]
+
+    await controller.cancel_all()
+
+    sent_commands = [
+        call.args[0] for call in controller.connection.send_command.call_args_list
+    ]
+    for index in controller.ramps:
+        assert f"N{index:02d}=0\r\n" in sent_commands
```
shihab-dls: I think this should do the checks on the attributes, not on the call arguments to `send_command`; check that `enabled` on each ramp has been set to false. This could be achieved simply by mocking out `enabled.put()`, and checking that this was called with `OnOffEnum.Off` for all ramps.

shihab-dls (approved):
Looks good, just left one comment on a test that I think should be implemented. Approving anyways as it is a nit.

tests/demo/test_controllers.py
coretl: Done in 8f3be18f — the test now mocks `enabled.put` on each ramp and asserts it was awaited once with `OnOffEnum.Off`, instead of matching `send_command` wire strings.

One note on the first half of your suggestion: I looked at asserting on the attribute *state* (`ramp.enabled.get() == OnOffEnum.Off`), but that can't work here — `enabled` has an `io_ref`, so `put` dispatches to `TemperatureControllerAttributeIO.send` and the readback only moves when the poll loop calls `update()`. Without a sim on the other end the value never changes, so mocking `put` (as you suggested) is the way to check this. Either way the assertion is now about what `cancel_all` is responsible for rather than the IO layer's formatting.

Also verified each ramp has its own `enabled` instance (not a shared class attribute), so all four assertions are real rather than one mock being checked four times.

---

Issue #394, opened 2026-07-20
ControllerFiller — declarative/procedural split


coretl:
## Scope
- `ControllerFiller` mirroring `ophyd_async` `DeviceFiller` (`core/_device_filler.py` — follow its **structure**; match names only where FastCS vocab fits). Scans class hints for `AttrR/W/RW[T]`, `Command[...]`, sub-`Controller`/`ControllerVector[T]`; creates children unfilled; `check_filled(source)` reports promised-but-missing.
- **Rule:** at the end of `__init__`, any Attribute referenced in code (i.e. type-hinted) MUST exist; `initialise()` may then run in parallel (only `__init__` is serial). (0013 L150)
- Support: **no-hints-at-all** (0013 L139); `Optional[X]` hints (0013 L153); adding attributes onto a **bare `Controller`** from outside — which is what fillers do (0013 L144).
- `yields (child, extras)` so third-party extras vocabularies work (decision 3); core defines none. Delete `HintedAttribute` + `_validate_type_hints`; remove the deepcopy half of `_bind_attrs` (keep `@command`/`@scan` binding). Trailing-underscore name convention.
- Migrate Example 5 (cut-down Eiger, #391) to the filler.

## ⚠️ Investigate
`fastcs-catio` builds Controller *classes* at runtime via `type(...)` from YAML before any instance exists — decide: does the filler support class-level dynamic construction, or must catio move to instance-level dynamic attrs? Record the decision in the PR / an ADR update.

## Key files (§9)
`controllers/base_controller.py` (`_bind_attrs`, `_find_type_hints`), + new filler module.

## Acceptance
`uv run --locked tox -e pre-commit,type-checking,tests` green; Example 5 updated + green.

ADR: 0013
Parent: #388
Blocked by: #392, #413

## Filler metadata validation — folded in from review
- Hinted children are created **unfilled in `__init__`** (so they exist after `__init__`, per the ADR 13 rule); `initialise()` only *fills* them (+ may add unhinted dynamic attrs).
- When filling from an `Annotated[Attr[T], extras]` hint, **runtime-validate** the metadata the extras carries (`FloatMeta`, or a protocol object's `.meta` e.g. `SCPIParam(...).meta`) against datatype `T` — `precision` on a `str` raises. Runtime counterpart to the static `Unpack[FloatMeta]` check (#392).

coretl:
[...]

PR #410, opened 2026-07-22
demo: cut-down Eiger REST sim + introspectable controller example

### LINKED ISSUE #391: Example 5 — cut-down Eiger: introspectable controller + REST sim
Part of the 5-example tutorial sequence (#388). **Step 5:** the introspectable (dynamic) device — the ladder's second backend, deliberately **not** the temperature sim.

Introspection earns its complexity only when a device's parameters aren't knowable at author time (a detector, not a fixed-command temp controller), so a separate Eiger REST backend gives the technique authentic motivation, exercises an HTTP-client backend the temp examples never touch, and matches real downstream drivers (`fastcs-eiger`, `fastcs-secop`, PandABlocks). The backend switch *is* the lesson: small & known → declare; large & self-describing → introspect.

## Scope
- Add `src/fastcs/demo/simulation/eiger.py`: a **FastAPI fake REST sim** with a few hardcoded, Eiger-shaped endpoints, introspectable (a `/keys` + per-key GET pattern). **No new dependency needed** — `fastapi` is already a core fastcs dependency and `httpx` is already a dev dependency (use FastAPI + `httpx`/`TestClient` in tests); do NOT run `uv add` (the nightly sandbox is offline).
- Add `src/fastcs/demo/eiger.py`: the FastCS side **type-hints half the endpoints (checked)** and **fills the other half by introspection**.
- **Baseline now (deliberately messy):** current introspection mechanism (`HintedAttribute`).
- **Clean form (when #394 lands):** `ControllerFiller` — migrated by #394.
- Unit tests against the fake REST sim (no hardware).

## Acceptance
`uv run --locked tox -e pre-commit,type-checking,tests` green.

ADR: 0013
Parent: #388
Blocked by: none

coretl:
Closes #391

Adds the Example 5 tutorial pieces from #388 §9 / the demo README ladder:

- `src/fastcs/demo/simulation/eiger.py` — a FastAPI fake REST sim shaped like a cut-down Eiger detector parameter tree: `config`/`status` subsystems, a `keys` listing endpoint per subsystem, and per-parameter GET/PUT. No new dependency (`fastapi[standard]` is already a core dep, `httpx` already a dev dep).
- `src/fastcs/demo/eiger.py` — an `EigerDetector` controller. Half its attributes (`count_time`, `state`) are declared as type hints and checked by the current `HintedAttribute` introspection-validation mechanism; the rest of the parameter tree is discovered at `initialise()` time by walking the sim's `keys` endpoints and added dynamically with no static check — exercising the current (baseline) introspection mechanism ahead of `ControllerFiller` (#394).
- `tests/demo/test_eiger.py` — unit tests against the fake sim (via `httpx.ASGITransport`, no real sockets/hardware): sim endpoint behaviour (list/get/put/read-only-rejection/404), hinted-vs-introspected attribute typing, and read/write round-trips through the controller.

Baseline uses the current API (`AttrR`/`AttrRW` + `io_ref`/`AttributeIO`); migrates to `ControllerFiller` when #394 lands, per the demo README.

1. `uv run pytest tests/demo/test_eiger.py -v`

### Checks for reviewer

### Notes
- Verified locally with `uv run --locked tox -e pre-commit,type-checking`, both green. For the `tests` env, this sandbox can't run `docs` (needs outbound network) or PVA-touching tests (needs a socket family unavailable here) — same known limitation noted on #409. Excluding those, `pytest src tests --ignore=tests/benchmarking` passes 316/326, with the same 10 pre-existing PVA/`p4p` failures (`RuntimeError: Address family not supported by protocol`), none related to this change. Real CI covers `docs` and PVA.

---

### DIFF
--- a/src/fastcs/demo/eiger.py
+++ b/src/fastcs/demo/eiger.py
@@ -27,6 +27,9 @@ _DATATYPES: dict[ValueType, type[DataType]] = {
     "bool": Bool,
 }
 
+# Poll period (seconds) for read-only status params that change on the device.
+UPDATE_PERIOD = 0.2
+
 
 @dataclass
 class EigerConnectionSettings:
@@ -106,6 +109,11 @@ class EigerDetector(Controller):
     count_time: AttrRW[float]
     state: AttrR[str]
 
+    # Derived (soft): built on top of the introspected ``state`` param. Declaring
+    # ``state`` as a checked attribute is what lets us reference it in code and
+    # publish something computed from it - here, whether the detector is idle.
+    idle = AttrR(Bool())
+
     def __init__(
         self,
         settings: EigerConnectionSettings | None = None,
@@ -129,11 +137,22 @@ class EigerDetector(Controller):
             for param in await self.connection.keys(subsystem):
                 data = await self.connection.get(subsystem, param)
                 datatype_cls = _DATATYPES[data["value_type"]]
-                io_ref = EigerAttributeIORef(subsystem=subsystem, param=param)
 
                 if data["access_mode"] == "rw":
+                    io_ref = EigerAttributeIORef(subsystem=subsystem, param=param)
                     attr = AttrRW(datatype_cls(), io_ref=io_ref)
                 else:
+                    # Read-only params are status values that change on the device,
+                    # so poll them periodically rather than reading once.
+                    io_ref = EigerAttributeIORef(
+                        subsystem=subsystem, param=param, update_period=UPDATE_PERIOD
+                    )
                     attr = AttrR(datatype_cls(), io_ref=io_ref)
 
                 self.add_attribute(param, attr)
+
+        # Keep the derived ``idle`` flag in sync with the introspected ``state``.
+        self.state.add_on_update_callback(self._update_idle)
+
+    async def _update_idle(self, state: str) -> None:
+        await self.idle.update(state == "idle")
--- a/src/fastcs/demo/eiger.py
+++ b/src/fastcs/demo/eiger.py
@@ -1,0 +1,139 @@
+"""Example 5 - introspectable controller: a cut-down Eiger over the fake REST sim.
+
+Half the attributes (``count_time``, ``state``) are declared as type hints and
+checked by the current ``HintedAttribute`` introspection-validation mechanism; the
+rest of the parameter tree is discovered at ``initialise()`` time by walking the
+sim's ``keys`` endpoints and is added dynamically, with no static check. A device
+that describes itself over the wire is exactly the case where introspection earns
+its complexity - contrast with the (deliberately non-introspectable) SCPI/temperature
+examples.
+"""
+
+from dataclasses import KW_ONLY, dataclass
+from typing import Any
+
+import httpx
+
+from fastcs.attributes import AnyAttributeIO, AttributeIO, AttributeIORef, AttrR, AttrRW
+from fastcs.controllers import Controller
+from fastcs.datatypes import Bool, DataType, Float, Int, String
+from fastcs.demo.simulation.eiger import API_PREFIX, Subsystem, ValueType
+from fastcs.util import ONCE
+
+_DATATYPES: dict[ValueType, type[DataType]] = {
+    "float": Float,
+    "int": Int,
+    "string": String,
+    "bool": Bool,
+}
+
+
+@dataclass
+class EigerConnectionSettings:
+    base_url: str = "http://localhost:8000"
+
+
+class EigerConnection:
+    """Thin async HTTP client wrapper for the Eiger REST sim.
+
+    A ``transport`` can be supplied to point directly at an in-process ASGI app
+    (e.g. in tests), bypassing the network entirely.
+    """
+
+    def __init__(self, transport: httpx.AsyncBaseTransport | None = None):
+        self._transport = transport
+        self._client: httpx.AsyncClient | None = None
+
+    async def connect(self, settings: EigerConnectionSettings) -> None:
+        self._client = httpx.AsyncClient(
+            base_url=settings.base_url, transport=self._transport
+        )
+
+    async def close(self) -> None:
+        if self._client is not None:
+            await self._client.aclose()
+            self._client = None
+
+    @property
+    def client(self) -> httpx.AsyncClient:
+        if self._client is None:
+            raise RuntimeError("EigerConnection is not connected")
+        return self._client
+
+    async def keys(self, subsystem: Subsystem) -> list[str]:
+        response = await self.client.get(f"{API_PREFIX}/{subsystem}/keys")
+        response.raise_for_status()
+        return response.json()
+
+    async def get(self, subsystem: Subsystem, param: str) -> dict:
+        response = await self.client.get(f"{API_PREFIX}/{subsystem}/{param}")
+        response.raise_for_status()
+        return response.json()
+
+    async def put(self, subsystem: Subsystem, param: str, value) -> None:
+        response = await self.client.put(
+            f"{API_PREFIX}/{subsystem}/{param}", json={"value": value}
+        )
+        response.raise_for_status()
+
+
+@dataclass
+class EigerAttributeIORef(AttributeIORef):
+    subsystem: Subsystem
+    param: str
+    _: KW_ONLY
+    update_period: float | None = ONCE
+
+
+class EigerAttributeIO(AttributeIO[Any, EigerAttributeIORef]):
+    def __init__(self, connection: EigerConnection):
+        super().__init__()
+        self._connection = connection
+
+    async def update(self, attr: AttrR[Any, EigerAttributeIORef]) -> None:
+        data = await self._connection.get(attr.io_ref.subsystem, attr.io_ref.param)
+        await attr.update(attr.dtype(data["value"]))
+
+    async def send(self, attr, value) -> None:
+        await self._connection.put(attr.io_ref.subsystem, attr.io_ref.param, value)
+
+
+class EigerDetector(Controller):
+    """Cut-down Eiger controller: half declared, half introspected."""
+
+    # Declared (checked): must exist, with this access mode and dtype, after
+    # initialise() introspects the parameter tree.
+    count_time: AttrRW[float]
+    state: AttrR[str]
+
+    def __init__(
+        self,
+        settings: EigerConnectionSettings | None = None,
+        transport: httpx.AsyncBaseTransport | None = None,
+    ) -> None:
+        self.connection = EigerConnection(transport=transport)
+        ios: list[AnyAttributeIO] = [EigerAttributeIO(self.connection)]
+        super().__init__(ios=ios)
+
+        self._settings = settings or EigerConnectionSettings()
+
+    async def connect(self) -> None:
+        await self.connection.connect(self._settings)
+        self._connected = True
+
+    async def disconnect(self) -> None:
+        await self.connection.close()
+
+    async def initialise(self) -> None:
+        for subsystem in ("config", "status"):
+            for param in await self.connection.keys(subsystem):
+                data = await self.connection.get(subsystem, param)
+                datatype_cls = _DATATYPES[data["value_type"]]
+                io_ref = EigerAttributeIORef(subsystem=subsystem, param=param)
+
+                if data["access_mode"] == "rw":
+                    attr = AttrRW(datatype_cls(), io_ref=io_ref)
+                else:
+                    attr = AttrR(datatype_cls(), io_ref=io_ref)
+
+                self.add_attribute(param, attr)
--- a/src/fastcs/demo/eiger.py
+++ b/src/fastcs/demo/eiger.py
@@ -9,14 +9,15 @@ its complexity - contrast with the (deliberately non-introspectable) SCPI/temper
 examples.
 """
 
+import enum
 from dataclasses import KW_ONLY, dataclass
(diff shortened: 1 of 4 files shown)


coretl:
Fixed the `docs / build` failure: `sphinx-build --fail-on-warning` was erroring on unresolvable autodoc cross-references to `httpx.AsyncBaseTransport` and `fastapi.applications.FastAPI` (no intersphinx mapping for either). Added both to `nitpick_ignore` in `docs/conf.py`, following the existing p4p precedent.

---

src/fastcs/demo/eiger.py
```diff
@@ -0,0 +1,139 @@
...
+        await self._connection.put(attr.io_ref.subsystem, attr.io_ref.param, value)
+
+
+class EigerDetector(Controller):
+    """Cut-down Eiger controller: half declared, half introspected."""
+
+    # Declared (checked): must exist, with this access mode and dtype, after
+    # initialise() introspects the parameter tree.
+    count_time: AttrRW[float]
+    state: AttrR[str]
```
coretl: The reason we put parameters on here is because we want to build code on top of them. Add an `idle: AttrR[bool]` attribute with value `self.state == "idle"`.

src/fastcs/demo/simulation/eiger.py
```diff
@@ -0,0 +1,91 @@
...
+    return {
+        "config": {
+            "count_time": EigerParameter(0.1, "float", "rw"),
+            "frame_time": EigerParameter(0.1, "float", "rw"),
+            "nimages": EigerParameter(1, "int", "rw"),
+            "description": EigerParameter("Simulated Eiger", "string", "r"),
+        },
+        "status": {
+            "state": EigerParameter("idle", "string", "r"),
+            "temperature": EigerParameter(22.5, "float", "r"),
```
coretl, line 54: Would be good to put a background task in here that oscillated between two temperatures, just so we can see something is updating on the front end

src/fastcs/demo/eiger.py
```diff
@@ -0,0 +1,139 @@
...
+        for subsystem in ("config", "status"):
+            for param in await self.connection.keys(subsystem):
+                data = await self.connection.get(subsystem, param)
+                datatype_cls = _DATATYPES[data["value_type"]]
+                io_ref = EigerAttributeIORef(subsystem=subsystem, param=param)
+
+                if data["access_mode"] == "rw":
+                    attr = AttrRW(datatype_cls(), io_ref=io_ref)
+                else:
+                    attr = AttrR(datatype_cls(), io_ref=io_ref)
```
coretl: Should add update_period to the read only params

src/fastcs/demo/eiger.py
coretl: Done in ff6292b0. Added a soft `idle = AttrR(Bool())` derived from `state`, kept in sync via an `add_on_update_callback` on `self.state` (`idle = state == "idle"`). Added a test covering it.

src/fastcs/demo/simulation/eiger.py
coretl: Done in ff6292b0. Added a lifespan background task (`_oscillate_temperature`) that sweeps `status/temperature` between two values with a sine so it visibly moves on the front end. It only runs under a real server's lifespan (uvicorn); the in-process ASGI transport used in tests doesn't start lifespan events, so the unit tests stay deterministic. Added a test that drives the lifespan explicitly to confirm it oscillates.

src/fastcs/demo/eiger.py
coretl: Done in ff6292b0. Read-only params now get an `update_period` (`UPDATE_PERIOD = 0.2`) so they poll; rw params still read once (`ONCE`). Added a test asserting the split.

tests/demo/test_eiger.py
```diff
@@ -0,0 +1,131 @@
...
+    response = await detector.connection.get("config", "count_time")
+    assert response["value"] == 0.5
+
+
+@pytest.mark.asyncio
+async def test_idle_derived_from_state(detector: EigerDetector):
+    # ``idle`` is soft and starts at its default, tracking ``state`` once polled.
+    assert detector.idle.get() is False
+
+    await detector.state.update("idle")
```
coretl: I think this is a hole in the current implementation, an `AttrR` should only allow `update(x)` if it doesn't have a getter. We can't put that guide in at the moment, but we should make the tests poke the sim via a backdoor, not update the attribute here

tests/demo/test_eiger.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+def test_sim_put_read_only_parameter_rejected(sim_client: TestClient):
+    response = sim_client.put(f"{API_PREFIX}/status/state", json={"value": "busy"})
+    assert response.status_code == 403
+
+
+def test_sim_unknown_parameter_404(sim_client: TestClient):
+    assert sim_client.get(f"{API_PREFIX}/config/nonexistent").status_code == 404
+    assert sim_client.get(f"{API_PREFIX}/nonexistent/keys").status_code == 404
```
coretl: sim tests aren't helpful, test the front door, the attributes

tests/demo/test_eiger.py
```diff
@@ -0,0 +1,131 @@
...
+    # bare ASGI transport used elsewhere, so drive the lifespan explicitly here.
+    app = create_eiger_sim_app()
+    transport = httpx.ASGITransport(app=app)
+    async with app.router.lifespan_context(app):
+        async with httpx.AsyncClient(base_url="http://sim", transport=transport) as c:
+            readings = []
+            for _ in range(4):
+                await asyncio.sleep(0.3)
+                response = await c.get(f"{API_PREFIX}/status/temperature")
+                readings.append(response.json()["value"])
```
coretl: again, test via the attributes, not the client. A good test for subscribe here.

src/fastcs/demo/simulation/eiger.py
```diff
@@ -0,0 +1,130 @@
...
+    """
+    mid = (low + high) / 2
+    amplitude = (high - low) / 2
+    start = time.monotonic()
+    while True:
+        elapsed = time.monotonic() - start
+        parameter.value = round(
+            mid + amplitude * math.sin(2 * math.pi * elapsed / period), 1
+        )
+        await asyncio.sleep(0.1)
```
coretl: This is overly complicated for a sim. A simple flip between 2 known values every 0.5s is sufficient. Makes the tests more predictable too

src/fastcs/demo/simulation/eiger.py
```diff
@@ -0,0 +1,130 @@
...
+
+    @asynccontextmanager
+    async def lifespan(app: FastAPI) -> AsyncIterator[None]:
+        task = asyncio.create_task(
+            _oscillate_temperature(state["status"]["temperature"])
+        )
+        try:
+            yield
+        finally:
+            task.cancel()
```
coretl, line 89: Can you use a FastAPI background task for this? Would that solve the lack of tasks in ASGI?

src/fastcs/demo/simulation/eiger.py
coretl: Done in da9988c5. Replaced the sine sweep with a simple flip between two known values (\`TEMPERATURES = (20.0, 30.0)\`) every 0.5s.

src/fastcs/demo/simulation/eiger.py
coretl: Kept it as the lifespan task. FastAPI's per-request \`BackgroundTasks\` are awaited *before* the response is sent, so a perpetual \`while True\` loop would hang the request - they're for short post-response work, not an app-lifetime task. The lifespan is the right tool (and is what uvicorn runs in a real server). For the ASGI gap: httpx's ASGITransport skips lifespan events, so the oscillation test now drives the lifespan explicitly (\`app.router.lifespan_context(app)\`) and observes the change through the controller attribute.

tests/demo/test_eiger.py
coretl: Done in da9988c5. Dropped the sim-only endpoint tests; the introspection/round-trip behaviour is now covered through the controller attributes instead. (The 404 / read-only-PUT-rejection paths aren't reachable through the controller front door - the controller never PUTs a read-only param - so they're gone.)

tests/demo/test_eiger.py
coretl: Done in da9988c5. Agreed on the \`AttrR.update\` hole. The test no longer calls \`update\` on the attribute: it pokes \`state\` through a sim backdoor (\`app.state.sim["status"]["state"].value = ...\`) and then polls the attribute via \`bind_update_callback\`, so it goes through the real IO path.

tests/demo/test_eiger.py
coretl: Done in da9988c5. Rewritten to go through the controller: builds an \`EigerDetector\`, subscribes to the temperature attribute with \`add_on_update_callback\`, polls across several sim flips, and asserts more than one distinct value was seen.

src/fastcs/demo/eiger.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+class EigerAttributeIO(AttributeIO[Any, EigerAttributeIORef]):
+    def __init__(self, connection: EigerConnection):
+        super().__init__()
+        self._connection = connection
+
+    async def update(self, attr: AttrR[Any, EigerAttributeIORef]) -> None:
+        data = await self._connection.get(attr.io_ref.subsystem, attr.io_ref.param)
+        await attr.update(attr.dtype(data["value"]))
```
shihab-dls: Here we're casting the value we recieve to the correct datatype before calling `update`. However, `attr.update` internally calls `datatype.validate()` which will try to cast the value anyway, and raise with a msg if it fails. We should not cast here in our `update` method, as the exception should be handled in one common place (i.e., `validate()`).

src/fastcs/demo/eiger.py
shihab-dls: We do not currently handle enum creation properly in fastcs-eiger, but I think we should assume that we do for the purposes of this demo. Therefore, for `state`, the actual response from the detector is: `{'access_mode': 'r', 'allowed_values': ['na', 'idle', 'ready', 'acquire', 'configure', 'initialize', 'error'], 'value': 'idle', 'value_type': 'string'}`, where we should use the `allowed_values` metadata to introspect an `AttrR[enum.Enum]` with the correct options. Can we amend the entry in the sim response, and accept `allowed_values` as a parameter on `EigerParameter`, then use this parameter in forming the client response from the server, such that we can correctly make an `AttrR[enum.Enum]` for `state`. I think we'll only need `[idle, ready, acquire]` for now. I'll un-resolve this thread to make my review comment visible.

tests/demo/test_eiger.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+@pytest_asyncio.fixture
+async def detector(_eiger) -> EigerDetector:
+    return _eiger[0]
+
+
+@pytest_asyncio.fixture
+async def sim(_eiger) -> SimState:
+    return _eiger[1]
```
shihab-dls, line 36: nit: usually I'd nitpick about putting the fixtures grouped at the top of the page, but in this case there is a very clear separation of concerns between the sim tests and detector tests, so I don't mind it as much.

shihab-dls (changes requested):
This looks really useful, I've left a few comments I want addressed before appoving this, so I'll request changes.

src/fastcs/demo/eiger.py
coretl: Done in 34196c6a. Dropped the cast - `update` now passes the raw device value straight to `attr.update`, which validates it. Agreed that coercion and the error message belong in one place.

This also turned out to be load-bearing for the enum change below: `Enum.validate` casts the wire string to the member by value, so the same `update` works for both.

tests/demo/test_eiger.py
coretl: Thanks - left the fixtures where they are then, for exactly that reason.

src/fastcs/demo/eiger.py
coretl: Done in 34196c6a.

- Sim: `EigerParameter` takes `allowed_values`, and GET includes it only for discrete params (as the real detector does). `state` now advertises `[idle, ready, acquire]`.
- Controller: a param that reports `allowed_values` is introspected as `Enum(enum.Enum(...))` built from those values, rather than `String`. `Enum.validate` casts the wire string to the member by value, so the IO's `update` needed no special-casing (it's the same `await attr.update(data["value"])` as everything else, now the cast is gone).
- `idle` is derived off `state.value == "idle"`.

One thing that fell out, worth flagging for #394: I couldn't keep a checked dtype on the hint. `state` is now `AttrR` (bare), not `AttrR[str]` or `AttrR[enum.Enum]`. `add_attribute` compares the hinted dtype for exact identity (`hint.dtype != attr.datatype.dtype`), so `AttrR[enum.Enum]` fails against the dynamically built subclass, and there's no author-time class to name. So the hint pins the access mode only. That's arguably the honest answer for a device-defined enum, but if `ControllerFiller` wants these checked, the hint check needs to accept a subclass (or grow a way to say "an enum, members TBD").

Tests: `state` is asserted to be an `Enum` datatype with `names == ["idle", "ready", "acquire"]`, plus a new `test_enum_attribute_reads_as_member` that pokes the sim backdoor and checks the attribute reads back as an enum member.

Issue #408, opened 2026-07-22
DOCS: Rewrite tutorials around the examples/ controllers (one per example, single-sourced)


coretl:
Rewrite the tutorials as a single pass once the `fastcs.demo` example controllers exist, sourcing all tutorial code directly from those living modules. Implements #388 §6.

## Decisions (settled with Tom, 2026-07-22)

1. **Single source of truth = `fastcs.demo`.** Tutorials `literalinclude` the finished example modules (consolidated into `src/fastcs/demo/`, mirroring ophyd-async); CI (`tox -e docs`) keeps them green so docs can't drift.
2. **Four tutorials** — a hello-world → complicated-device ladder (five example modules; the old "reusable `io=` object" rung is gone, because `io=` objects were replaced by getter/setter callables):

   | # | Tutorial | Teaches | Backend | Module(s) |
   |---|----------|---------|---------|-----------|
   | 1 | hello world | pure-soft `@attr` decorator | none | `hello_world.py` (#398) |
   | 2 | getter/setter | `getter`/`setter` callables in `__init__`; closes with "for a real family, go declarative →" | temperature | `temperature_attr.py` (#404) |
   | 3 | declarative | annotated `SCPIParam` + filler (static metadata) **plus** composition — sub-controllers/vector + `@scan` + `@command` on the full temp controller | temperature | `temperature_scpi.py` + `scpi.py` (#405), drawing on `controllers.py` (#390) |
   | 4 | introspectable | bare hints filled from a **runtime** REST parameter tree | Eiger REST | `eiger.py` (#391) |

3. **Finished-example walkthrough** pedagogy (not the old 15-diff "Code N" build). Retire `docs/snippets/static01–15.py` and `docs/snippets/dynamic.py`.
4. **One umbrella docs pass**, after the examples land (see `Blocked by:`).

## Scope
- Four tutorial pages replacing the current `static-drivers` / `dynamic-drivers` split; update `tutorials.md` nav; keep `installation.md`.
- All code blocks come from the `fastcs.demo` modules via `literalinclude` with region markers (`:start-after:` / `:end-before:`). **Add the markers to each module as part of writing its tutorial here.**
- **State the declarative (class-body) / procedural (instance-scope) rule up front**, mirroring ophyd-async's `declarative-vs-procedural` page.
[...]

PR #411, opened 2026-07-22
demo: convert temperature controller to getter/setter style

### LINKED ISSUE #404: Example 2 — attr getter/setter via __init__ (temperature backend)
Part of the tutorial examples (#388). **Tutorial 2 — getter/setter:** per-attribute IO as `getter`/`setter` callables supplied in `__init__` (`AttrRW(getter=…, setter=…)`), on the **temperature-controller backend**.

Retargeted from a standalone Thorlabs MFF device onto the temperature sim (per Tom, 2026-07-22) so the ladder keeps to two backends.

## Scope
- Add `src/fastcs/demo/temperature_attr.py`: a small temperature controller (e.g. `ramp_rate`, `power`) reusing the temperature sim; no bespoke device.
- **Baseline now (current callback-IO API) — use the single callable-wrapping IO + protocol shape, NOT one IO subclass per attribute:**
  - a **single** generic `TemperatureIO(AttributeIO[...])` registered once (`ios=[TemperatureIO(conn)]`) whose `update`/`send` just invoke callables carried on the ref;
  - a `TemperatureIORef(AttributeIORef)` carrying the command-building callables (`read_cmd: Callable[[], str]`, `write_cmd: Callable[[float], str] | None`);
  - a single `TemperatureProtocol` class with one method per command (the getters/setters), referenced by each attribute's ref — mirrors `fastcs-thorlabs-mff` (`MFFAttributeIO`/`MFFAttributeIORef`/`ThorlabsAPTProtocol`).
  - Response parsing stays inline in `TemperatureIO.update` (`float(...)`), not a separate `response_handler` callable.
  - This is the honest precursor to #392: `read_cmd`/`write_cmd` **are** the getter/setter; #392 promotes them onto the constructor and deletes the IO/ref wrapper, while `TemperatureProtocol` survives unchanged.
- **Clean form (when #392 lands):** `AttrRW(Float(), getter=protocol.get_…, setter=protocol.set_…)` (getter/setter constructor params; datatype inferred from the getter's return annotation).
- Unit tests: read/write against the sim.

## Notes
- The tutorial closes by pointing at the declarative style (#405) as the DRY answer *when the shared wire pattern is worth naming* — not "more than N attributes".
- The `@attr` decorator form (#397) is the sugar over this same getter/setter; #398 (hello-world) is the decorator's showcase.

## Acceptance
`uv run --locked tox -e pre-commit,type-checking,tests` green (full tox in real CI on the PR).

ADR: 0014 (getter/setter), 0018 (`@attr`)
Parent: #388
Blocked by: none

coretl:
Closes #404

Retargets Tutorial 2 (getter/setter) onto the **existing** temperature controller demo rather than adding a second module: the existing `src/fastcs/demo/controllers.py` is converted to the getter/setter style in place and renamed to `src/fastcs/demo/temperature_attr.py`, matching how the other demo modules are named (device + style, not framework concept). One temperature backend, one temperature module.

Current-API baseline in the intended shape: `TemperatureProtocol` / `TemperatureRampProtocol` carry one method per wire command, each attribute's `TemperatureIORef` carries the command-building callables (`read_cmd`/`write_cmd`), and a **single** generic `TemperatureIO` just invokes them — mirroring `fastcs-thorlabs-mff` (`MFFAttributeIO`/`MFFAttributeIORef`/`ThorlabsAPTProtocol`). `read_cmd`/`write_cmd` **are** the getter/setter this foreshadows: #392 promotes them onto `AttrRW(getter=..., setter=...)` and deletes the IO/ref wrapper, while the protocol classes survive unchanged. Response parsing (`attr.dtype(...)`) is inline in `TemperatureIO.update`, not a `response_handler` callable.

Attributes move from the class body into `__init__`. That is what earns the style: each ramp bakes its index into its own protocol instance, instead of the shared IO dispatching on a `name` string plus a suffix. The composition and methods rungs are carried through unchanged — the `ControllerVector` of `TemperatureRampController`s from #409, `@scan` and `@command` — so this module now covers both the getter/setter rung and the composition rung (#390), and the README ladder collapses from five modules to four accordingly.

Behaviour is preserved: same attribute names and hierarchy, same 0.2 s default poll period, same two scan groups. The launch `type:` in `fastcs.yaml` is derived from the top-level package rather than the submodule, so `fastcs.TemperatureController` and the checked-in `schema.json` survive the rename unchanged (verified by regenerating the schema).

1. `uv run pytest tests/demo/test_temperature_attr.py -v`
2. `python -m fastcs.demo run src/fastcs/demo/fastcs.yaml` against the sim (`tickit all src/fastcs/demo/simulation/temp_controller.yaml`)

### Checks for reviewer
- [ ] Tutorial 2 now opens on the full multi-ramp controller rather than a two-attribute toy. If it should stay minimal, the fix is at docs time (#408): have the tutorial `literalinclude` only the `ramp_rate`/`power` region instead of the whole module.

### Notes
- Rebased onto `refactor` after #409/#410 merged; the `ControllerVector` ramps and the existing `tests/demo/test_controllers.py` cases (`test_ramps_is_controller_vector`, `cancel_all`, `update_voltages`) are carried into the rewrite and the renamed test module, alongside the new wire-level protocol tests.
- Verified locally with `uv run --locked tox -e pre-commit,type-checking`, both green.
- Also smoke-tested end-to-end against the real tickit sim: float write+read (`ramp_rate`), int write+read (`end`), enum write (`Started ramp 0` in the sim log, `target` → 10.0), read-only reads, `update_voltages` and `cancel_all`.
- As on #409/#410, this sandbox can't run the `docs` env (no outbound network) or PVA-touching tests (no PVA-capable socket family). Excluding those and Tango (the sandbox cannot start a Tango device server — `psutil.NoSuchProcess`, reproducible on `refactor` itself), `pytest src tests --ignore=tests/benchmarking` gives 282 passed. Real CI covers `docs`, PVA and Tango.

---


src/fastcs/demo/temperature_attr.py
```diff
@@ -0,0 +1,104 @@
...
+        super().__init__(ios=[TemperatureIO(self.connection)])
+
+        self.ramp_rate = AttrRW(
+            Float(),
+            io_ref=TemperatureIORef(
+                read_cmd=self._protocol.get_ramp_rate,
+                write_cmd=self._protocol.set_ramp_rate,
+                update_period=0.2,
+            ),
+        )
```
coretl, line 161: This is much more like the MFF. I think this is the right "messy place" to start iterating from.

tests/demo/test_temperature_attr.py
```diff
@@ -0,0 +1,48 @@
...
+        ip_settings=IPConnectionSettings(ip="localhost", port=25565)
+    )
+    controller = TemperatureAttrController(settings)
+    controller.post_initialise()
+    return controller
+
+
+@pytest.mark.asyncio
+async def test_ramp_rate_read_from_device(controller: TemperatureAttrController):
+    controller.connection.send_query = AsyncMock(return_value="1.5\r\n")  # type: ignore[method-assign]
```
coretl: @shihab-dls is this how we expect people to be writing tests? Or is there another "mocked IP connection" object we should be using?

shihab-dls: Pretty much, yeah. On `fastcs-odin` unit tests, for example, we just
```python
controller = OdinController(OdinControllerSettings(IPConnectionSettings("", 0)))
controller.connection = mocker.AsyncMock()
```
and mock responses directly, like we do here. In both of these cases, we are making a `Settings` object, which we could just pass as a `MagicMock()` anyway, but It makes no real difference to doing `IPConnectionSettings(ip="localhost", port=25565)`.

shihab-dls: I think linting still passes without `# type: ignore[method-assign]`.. Does this have to do with eventually turning strict mode on?

shihab-dls (approved):
Just one comment regarding `# type: ignore`. This is idiomatic fastcs logic, with good tests, and It does show obscurity with passing setters and getters into an `AttributeIORef`, which I think is a good starting point. Approving.

tests/demo/test_temperature_attr.py
coretl: Good catch - removed in 1051a645. They were leftover mypy-isms: this project type-checks with pyright (`standard` mode) and `method-assign` is a mypy error code pyright never emits, so the comments were pure noise rather than strict-mode groundwork. `pyright src tests` is clean without them and the three tests still pass.

Issue #395, opened 2026-07-20
AttrW setpoint cache, native timestamps + severity, ControllerRunner


coretl:
## Scope
- **AttrW setpoint cache** with a `.get_setpoint()` method (0016 L96) — needed for ophyd `locate()`, useful to all transports.
- **Native timestamps (+ severity)** on attribute updates: follow the `Reading` shape but share **no** code (0016 L101); severity is a **fastcs enum using the same strings as EPICS** (0016 L104).
- **`ControllerRunner`**: a class with `start`/`stop` (context manager only if it makes sense for the `FastCS()` case; ophyd-async uses `start`/`stop`; idempotency is the caller's responsibility). It **owns the whole lifecycle incl. reconnect** (0016 L108/L112). Extract from `FastCS.serve`. Document the stable interface (decision 13).

## Key files (§9)
`src/fastcs/attributes/`, `control_system.py` (`FastCS.serve`).

## Acceptance
`uv run --locked tox -e pre-commit,type-checking,tests` green.

ADR: 0016
Parent: #388
Blocked by: #392

coretl:
Opened #420 for this.

`ControllerRunner` is extracted from `FastCS.serve` with `setup`/`start`/`stop`, native timestamps and severity ride on `Update`, and the stable interface is written down in `docs/explanations/stable-interface.md`.

Two things worth knowing:

- **The setpoint cache this issue lists was already delivered by #412**, as the `.setpoint` property — cached by `set()` before the setter runs and regardless of whether it succeeds, which is what ADR 0016's question 1 settled on. It is documented as part of the stable surface here, but there was no code to write. (The issue text says `.get_setpoint()`; ADR 0014's runtime-surface rename made it a property, and ADR 0016 Q1 agrees.)
- **Reconnect turns out to have been dead.** A scan task that raises sets `_connected = False` and pauses, but nothing in FastCS ever called `Controller.reconnect()` — so a controller that dropped out stayed paused unless its driver wired its own recovery. The runner now owns that, per the ADR, which is a behaviour change rather than a pure move; flagged for review on the PR.

---
_Generated by [Claude Code](https://claude.ai/code)_

coretl:
Closed by #420, merged into `refactor`. (GitHub only auto-closes linked issues for PRs merged into the default branch.)

— overnight agent

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #397, opened 2026-07-20
@attr decorator sugar over getter/setter constructors


coretl:
Adds the `@attr` decorator sugar over the `getter=`/`setter=` constructors from #392 (ADR 0018, 0014 review update).

## Scope
- `@attr` is a **decorator only**, two forms: bare `@attr` and parameterised `@attr(precision=3, units="V", update_period=0.5)` on an async getter, plus property-style `@my_attr.setter` on the writer.
  - getter only → `AttrR`; getter + `.setter` → `AttrRW` (deferred creation). There is **no** `@attr_r`/`@attr_rw`.
  - **No free-function `attr()` factory** — the procedural spelling is `AttrR(getter=…)` / `AttrRW(getter=…, setter=…)` (from #392); `@attr` is pure sugar over it.
  - `AttrW`-only (write, no getter) is rare → written longhand as `AttrW(setter=…)`, no decorator.
- Datatype inferred from the getter's return annotation (unwrapping `Update[T]`); decorator kwargs typed `Unpack[*Meta]`, validated against it. Getter docstring → attribute `description`.
- Introspected vs decorated attrs: don't shadow already-defined ones; raise on name clash.
- Built on the existing `Unbound*` bind machinery (fresh objects per instance, no prototype/deepcopy). Include the "FastCS for PyTango users" docs page.
- **Unblocks #398** (hello-world uses `@attr`); #404's `__init__` getter/setter can also be shown in `@attr` form for contrast.

## Key files (§9)
`src/fastcs/methods/` (`Unbound*` bind machinery), `attributes/`.

## Acceptance
`uv run --locked tox -e pre-commit,type-checking,tests` green; #398 (hello-world) green against `@attr`.

ADR: 0018, 0014
Parent: #388
Blocked by: #392, #413

coretl:
Implemented in #423 (based on `refactor`).

`@attr` + `@x.setter` over the `getter=`/`setter=` constructors, bound per instance through an `UnboundAttr` declaration the same way `@command`/`@scan` are. Datatype from the getter's return annotation, docstring summary as the description, decorator kwargs as `Unpack[Meta]` validated against the datatype, and a `Polled`/`NotPolled` schedule as the leading positional. Plus the "FastCS for PyTango users" docs page.

[...]

PR #412, opened 2026-07-23
attributes: getter/setter IO rework, remove AttributeIORef/AttributeIO

### LINKED ISSUE #392: getter/setter IO rework; remove AttributeIORef and AttributeIO

coretl:
Closes #392

Implements the getter/setter half of #392 (ADR 0014's `AttributeIO`/`AttributeIORef` removal), scoped per the issue's own "Note on size": the `DataType`/`*Meta` TypedDict replacement and the precision/`Limits` naming pass (ADR 0017) are left for a follow-up PR rather than risking an incoherent halfway state in one session — see "Notes" below.

### Scope

- **Per-attribute IO is now `getter`/`setter` callables** passed straight to `AttrR(getter=...)`/`AttrW(setter=...)`/`AttrRW(getter=..., setter=...)`. Access mode is enforced structurally by which of `getter`/`setter` are present - there's no more `io=`/`AttributeIO` class hierarchy or `AttributeIORef`-keyed dispatch registry.
- **Datatype is now optional** on the constructors when it can be inferred from the getter's return annotation or the setter's value-parameter annotation (`AttrR(getter=get_value)` where `get_value() -> float`); not inferable ⇒ positional datatype still required, fails fast at construction.
- **New `Update[T]`** dataclass (`value`, `timestamp`) a getter/setter may return/accept instead of a bare value - `update()`/`poll()` unwrap it. (Native persistence of the timestamp, and a `severity` field, are left to #395, which explicitly owns that.)
- **Runtime surface rename** (ADR 0014): `get()` → `.readback`/`.setpoint` read-only properties (`AttrR` has `readback`, `AttrW` has `setpoint`, `AttrRW` has both); no-arg `update()` → `poll()` (does the getter read, caches, returns the value); `update(value)` stays as a pure cache-push (now also accepting `Update[T]`); `put(value[, sync_setpoint])` → `set(value)` (caches `.setpoint`, runs the setter; a non-`None` setter return updates `.readback` too - the sanctioned replacement for the old private `_call_sync_setpoint_callbacks`/`sync_setpoint=` mechanism).
- **Removed**: `AttributeIO`, `AttributeIORef`, `ios=`, `_connect_attribute_ios`, `_validate_io`, `_attribute_ref_io_map`, `set_update_callback`/`bind_update_callback`/`set_on_put_callback`/`_call_sync_setpoint_callbacks`/`add_sync_setpoint_callback`, and the second `Attribute`/`AttrR`/`AttrW`/`AttrRW` TypeVar (`Attribute[DType_T, AttributeIORefT]` → `Attribute[DType_T]`).
- `Controller.create_api_and_tasks` now schedules polling directly off getter-bearing attributes (`poll_period`, defaulting to `ONCE` when a getter is given) instead of pattern-matching an `AttributeIORef`.
- Migrated: the demo composition example (`src/fastcs/demo/controllers.py`), every `docs/snippets/*.py` tutorial snippet that used `io_ref=`, and all the tests that constructed attributes the old way (`tests/assertable_controller.py`, `tests/test_attributes.py`, `tests/test_control_system.py`, `tests/conftest.py`, `tests/example_p4p_ioc.py`, and the transport test files that built attributes directly).
- `.datatype`/`DataType` themselves are untouched in this PR (see Notes).

### Notes

- **Deliberately out of scope, left for a follow-up**: the `DataType` family → python type + `*Meta` TypedDict replacement and the `precision`/nested-`Limits` naming pass (ADR 0017). The issue's own "Note on size" calls out this split explicitly. `attr.datatype`/`Float`/`Int`/etc. are unchanged.
- Verified locally with `uv run --locked tox -e pre-commit,type-checking`, both green in full. For the `tests` env, this sandbox can't run `docs` (needs outbound network to `diamondlightsource.github.io`) or the PVA/p4p-backed tests (`RuntimeError: Address family not supported by protocol` - no PVA-capable socket family here), the same known limitation noted on prior PRs (#409/#410/#411) in this epic. Excluding those two paths, `pytest src tests --ignore=tests/benchmarking` passes 314/324, with only the same 10 pre-existing PVA (`p4p`)/socket-family failures, unrelated to this change. Real CI covers `docs` and PVA.
- Also spot-checked the demo end-to-end against the real `tickit` temperature-controller simulator (`docs/snippets/*.py`'s `test_docs_snippets.py` already exercises this - all 16 snippets pass) and confirmed `TemperatureController`/`TemperatureRampController` construct and wire up correctly with the new getter/setter API.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TpF6ugRp1Z64YPNQ1WoZUT)_


src/fastcs/attributes/attr_r.py
```diff
@@ -1,75 +1,101 @@
...
+class AttrR(Attribute[DType_T]):
     """A read-only ``Attribute``"""
 
     def __init__(
         self,
-        datatype: DataType[DType_T],
-        io_ref: AttributeIORefT | None = None,
+        datatype: DataType[DType_T] | None = None,
+        getter: Getter[DType_T] | None = None,
+        poll_period: float | None = _UNSET,  # type: ignore[assignment]
```
coretl: Merge these 2 arguments, so we can decorate `getter` with the update period. E.g.

```python
attr = AttrR(getter=Polled(protocol.get_temperature, period=0.1))
```

src/fastcs/attributes/attr_r.py
coretl, line 1: There should be symmetry in the callbacks:
- `add_readback_callack` (was `add_on_update_callback`) that is called whenever the readback updates
- `add_setpoint_callback` that is called whenever the setpoint updates
The transports should never update their own sepoints, but should do so via the `add_setpoint_callback`.

For the `sync_setpoint`, this should vanish, and be replace with an addition to `Update`. Update should now have:
```python
@dataclass
class Update(Generic[T]):
    readback: T
    timestamp: float | None = field(default_factory=time.time)
    severity: Severity = Severity.NO_ALARM
    setpoint: T | None = None  # If None then don't update setpoint, otherwise update to this value
```

docs/explanations/transports.md
```diff
@@ -140,49 +139,55 @@ def create_read(name, attribute):
...
     protocol_setpoint = Protocol(name)
 
     async def handle_write(value):
         protocol_setpoint.post(value)
-        await attribute.put(value)
+        await attribute.set(value)
 ```
 
-### Sync Setpoint Callbacks
+### Seeding a Setpoint Display from the Readback
```
coretl: I think this can be simplified a lot. If we say that each transport should update its setpoint *every time* the Attr changes its setpoint, then we can ditch the idea of seeding, and instead say that transports should `add_setpoint_callback` with something that updates their setpoint. Make this change.

This means that all transports will show the same setpoint, which is actually desirable for consistency. Working out which transport did the update can be done from logs. This needs an ADR as it is a decision.

src/fastcs/attributes/attr_r.py
coretl: Need to consider what happens if we have a getter that returns a `T` rather than an `Update[T]`. 

I think the correct behaviour should be at the Attr level, start without a setpoint, then the first `update` or `poll` we set setpoint=readback

src/fastcs/attributes/attr_w.py
```diff
@@ -1,98 +1,80 @@
...
-        """Callbacks to publish changes to the setpoint of the attribute"""
+        if datatype is None:
+            datatype = setter and infer_datatype_from_setter(setter)
+            if datatype is None:
+                raise ValueError(
+                    "datatype must be given explicitly, or be inferable from the "
+                    "setter's value parameter annotation"
+                )
+
+        Attribute.__init__(self, datatype, group, description=description)
```
coretl: Why not super init?

src/fastcs/attributes/attr_rw.py
```diff
@@ -1,42 +1,79 @@
...
+                    "datatype must be given explicitly, or be inferable from the "
+                    "getter/setter annotations"
+                )
 
-        if io_ref is None:
-            self.set_on_put_callback(self._internal_update)
+        AttrR.__init__(
+            self, datatype, getter, poll_period, group, initial_value, description
+        )
+        AttrW.__init__(self, datatype, setter, group, description)
```
coretl: code smell: AttrR takes datatype possibly from getter and passes to Attribute. AttrW takes datatype possibly from setter and passes to Attribute. AttrRW has to do both to pass it down to both. Is there a better way?

coretl: Possible: make AttrR and AttrW allow **kwargs and pass them down, then make Attribute strict. Then we don't need this datatype handling as either AttrR or AttrW will have calculated it (and passed it through) so Attribute gets it

src/fastcs/attributes/attr_r.py
coretl: Done - `Polled` in `attributes/attr_r.py`, exported from `fastcs.attributes`:

```python
self.ramp_rate = AttrRW(Float(), getter=Polled(protocol.get_ramp_rate, period=0.2))
```

`poll_period` is gone from the constructors. The three schedules are now spelled by how the getter is passed: a bare getter is `ONCE` (still the default when a getter is given), `Polled(getter, period=0.2)` polls, and `Polled(getter, period=None)` is never scheduled - on-demand `poll()` only. `period` is deliberately required on `Polled` rather than defaulting, so there is no arbitrary magic number.

src/fastcs/attributes/attr_r.py
coretl: Done, both parts.

**Callbacks** are now symmetric: `add_on_update_callback` -> `AttrR.add_readback_callback`, and a new `AttrW.add_setpoint_callback` beside it. Both EPICS transports register a setpoint callback instead of driving their own setpoint display - see the transports.md thread and ADR 0020.

**`sync_setpoint` is gone**, replaced by `Update`:

```python
@dataclass
class Update(Generic[DType_T]):
    readback: DType_T
    timestamp: float | None = None
    setpoint: DType_T | None = None  # None -> leave the cached setpoint alone
```

A bare value returned from a setter still means both (accepted/clamped -> readback *and* setpoint); an explicit `Update` splits them.

Two deliberate deviations from your snippet, both easy to change if you disagree:

- **No `severity` field.** #395 explicitly owns native timestamps + severity, and a `Severity` enum with nothing consuming it is dead code until the transports plumb it into alarm states. Happy to add it here if you would rather have the shape settled now.
- **`timestamp: float | None = None`** rather than `field(default_factory=time.time)`. Those are different behaviours: `None` means "framework stamps receive-time", which is what ADR 0014 specifies, whereas a default factory stamps *construction* time. Persisting the timestamp natively is #395 too.

coretl: Done - implemented exactly as you describe. An `AttrRW` starts with no known setpoint (`_setpoint_known = False`; the cache holds the datatype default but means nothing yet), and `AttrRW.update` mirrors readback -> setpoint the first time a value arrives from anywhere - a poll, a scan, any `update()`:

```python
async def update(self, value):
    await super().update(value)
    if isinstance(value, Update) and value.setpoint is not None:
        await self.update_setpoint(value.setpoint)
    elif not self._setpoint_known:
        await self.update_setpoint(self._value)
```

Only the *first* readback does this. Later ones do not, so a readback that disagrees with the setpoint does not silently rewrite what the user asked for. A getter returning `Update(readback=..., setpoint=...)` drives the setpoint explicitly instead, for devices that report their own.

This is what let the transport-side seeding go away entirely.

docs/explanations/transports.md
coretl: Done, and it did simplify a lot.

`AttrW` gained `add_setpoint_callback()` / `update_setpoint()`. `set()` caches and publishes the requested value *before* running the setter, then publishes again if the setter returns a clamped value. Both EPICS transports now just register a callback; the two near-identical one-shot seeding blocks and their `isinstance(attribute, AttrR)` guards are deleted, and the mechanism works for a pure `AttrW` too, which seeding never did.

Written up as **ADR 0020** (`docs/explanations/decisions/0020-transport-setpoint-mirroring.md`), including the consequence you called out: all transports show the same setpoint, attribution is deliberately traded away, and the logs are where you recover which transport wrote. It also picks up the CA-vs-PVA ordering divergence from the #402 review - CA used to post the setpoint only after the update callback, so a slow setter left it stale; both now post at the start of the write, which was PVA behaviour.

transports.md is rewritten to match: new "Setpoint Callbacks" section listing the four things that fire it, "Seeding a Setpoint Display from the Readback" deleted.

src/fastcs/attributes/attr_w.py
coretl: Done - `super().__init__()` throughout. The explicit `Attribute.__init__` calls were there to dodge the `AttrRW` MRO, which is the same root cause as the datatype smell you spotted below; fixing that one made cooperative `super()` work, so this fell out with it.

src/fastcs/attributes/attr_rw.py
coretl: Agreed, and your second suggestion is what I implemented.

`AttrR` infers from the getter, `AttrW` from the setter, and each passes the (possibly still unresolved) datatype down via `**kwargs`. The MRO `AttrRW -> AttrR -> AttrW -> Attribute` runs both in turn, so whichever *can* resolve it does, and `Attribute` is now strict - it raises if the datatype is still `None` when it arrives.

`AttrRW.__init__` has no datatype handling left at all:

```python
super().__init__(datatype, getter=getter, setter=setter, initial_value=initial_value, **kwargs)
```

The error message also now exists once instead of three times, and `Attribute` is the single place that decides a datatype is missing.

src/fastcs/attributes/attr_r.py
coretl: Follow-up — this evolved past the `Polled(getter, period=…)` I first replied with. Final shape, after discussing the asymmetry with the `@attr` decorator:

| Schedule | Procedural | Declarative (#397) |
|---|---|---|
| Once, at connect | `AttrR(t, getter=g)` | `@attr(units="V")` |
| Every 0.5s | `AttrR(t, getter=Polled(g, period=0.5))` | `@attr(Polled(0.5), units="V")` |
| Never; `poll()` only | `AttrR(t, getter=NotPolled(g))` | `@attr(NotPolled(), units="V")` |
| Soft (no IO) | `AttrR(t)` | — |

Three changes from my first reply:

- **`NotPolled(getter)` replaces `Polled(getter, period=None)`.** A `Polled` that is never polled was a contradiction, and the two differed only by an argument being present.
- **`period` is keyword-only**, so a period always says what it is.
- **Both wrappers take an optional getter and bind one when called**, so `Polled(period=0.5)(g)` works. That is what lets the same objects serve `@attr`, where the getter arrives by decoration and there is no argument to wrap — the decorator takes a schedule positionally rather than a `poll_period=` kwarg the constructor no longer has.

A bare getter stays **read-once-at-connect** rather than becoming unpolled. That was the one point I argued against: a bare `@attr` has to resolve to some schedule (ADR 18), so a constructor that refused to default while the decorator defaulted would put the asymmetry straight back. Of the two candidate defaults, once-at-connect is the one that fails safe — unpolled-by-default leaves an `AttrRW` at the datatype default, and under ADR 20 it would then never establish a setpoint either, so every transport would show `0`/`""`/`False` until someone wrote to it.

ADR 0014 has been **rewritten** rather than amended (the refactor-branch ADRs are unreleased): the schedule now has its own section with the table above and the three candidate defaults, `Update` is documented as built (`readback`/`timestamp`/`setpoint`; `severity` is ADR 16's), the runtime-surface table lists `update_setpoint()` and both callback registrars, and question 6 — whether the setpoint echo is visible across transports — is now answered rather than deferred, since ADR 20 closed the CA-lags-PVA follow-up it left open. ADR 0018 is amended to match, and 0013's stale `poll_period=` example is fixed.

src/fastcs/attributes/attr_r.py
```diff
@@ -85,57 +159,45 @@ async def update(self, value: Any) -> None:
...
-        """Set the callback to update the value of the attribute from the source
+    def add_readback_callback(
+        self, callback: AttrReadbackCallback[DType_T], always: bool = False
+    ) -> None:
+        """Add a callback to be called when the readback of the attribute updates
 
-        The callback will be converted to an async task and called periodically.
+        The callback will be called with the updated readback value. Transports
+        should use this to publish the attribute's readback, and
+        ``AttrW.add_setpoint_callback`` to publish its setpoint.
```
shihab-dls, line 195: should: the `always` parameter determines if the callback is called even if the new value is equal to the current cached value. A description of this should be included in the docstring, as this is only inferrable from operations in `update()`.

src/fastcs/attributes/attr_rw.py
```diff
@@ -1,42 +1,84 @@
...
+        elif not self._setpoint_known:
+            await self.update_setpoint(self._value)
+
+    async def set(self, value: DType_T) -> None:
+        """Request a new value for the attribute.
+
+        With no setter, this is a soft attribute: the requested value is pushed
+        straight to the readback. With a setter, a returned value is additionally
+        applied to the readback - the sanctioned replacement for the old private
+        setpoint-echo mechanism.
```
shihab-dls: nit: I don't see why "the sanctioned replacement for the old private setpoint-echo mechanism" should be included in the docstring for `set()`. This seems like the sort of sentence you'd see in an ADR, not in a method docstring.

tests/example_p4p_ioc.py
```diff
@@ -33,39 +21,30 @@ class FEnum(enum.Enum):
...
     @command()
     async def d(self):
         print("D: RUNNING")
         await asyncio.sleep(0.1)
         print("D: FINISHED")
-        await self.j.update(self.j.get() + 1)
+        await self.j.update(self.j.readback + 1)
 
-    e: AttrR = AttrR(Bool(), io_ref=SimpleAttributeIORef())
+    e: AttrR = AttrR(Bool())
```
shihab-dls, line 60: Should: we've gotten rid of the IOs, but have not replaced them with anything. Add an attribute that uses a simple setter.

tests/test_attributes.py
```diff
@@ -47,34 +44,116 @@ def test_attr_r():
...
+
+@pytest.mark.asyncio
+async def test_poll_with_no_getter_raises():
+    attr = AttrR(Int())
 
-    await attr.update("200")  # type: ignore
-    assert attr.get() == 200
-    sync_setpoint_mock.assert_called_once_with(200)
+    with pytest.raises(RuntimeError):
+        await attr.poll()
```
shihab-dls: Should: we shouldn't be checking against only the exception type. Match this with the full or subset of the exception message.

tests/test_attributes.py
```diff
@@ -47,34 +44,116 @@ def test_attr_r():
...
+
+@pytest.mark.asyncio
+async def test_poll_exception_propagates():
+    async def do_update():
+        raise ValueError("do_update failed")
+
+    attr = AttrR(Int(), getter=do_update)
+
+    with pytest.raises(ValueError):
+        await attr.poll()
```
shihab-dls: Should: same here, we should match with the expected exception message or a subset of it.

src/fastcs/attributes/attr_r.py
coretl: Done in ec2f44c4 — `always` now has an `Args:` entry saying it controls whether the callback fires on every `update()` rather than only when the value changes, and what it is for (a callback that timestamps or counts updates rather than displaying them).

---

src/fastcs/attributes/attr_rw.py
coretl: Agreed — dropped in ec2f44c4. That sentence was justifying the design against what it replaced, which belongs in ADR 0014/0020 where it already is. The docstring now just states the behaviour: a returned value is the device's accepted/clamped value and is applied to the readback as well as the setpoint.

---

tests/example_p4p_ioc.py
coretl: Good catch — this IOC lost all of its IO when `AttributeIO` went, so nothing in it exercised the replacement by hand any more.

Added in ec2f44c4: `ChildController.clamped`, an `AttrRW` with a getter/setter pair over an in-memory value. The setter clamps to 0..100 and returns what it accepted:

```python
async def get_clamped(self) -> int:
    return self._clamped

async def set_clamped(self, value: int) -> int:
    self._clamped = min(max(value, 0), 100)
    return self._clamped
```

I picked a *clamping* setter rather than a plain one so the IOC demonstrates both halves of ADR 0020 interactively: the getter seeds the setpoint at connect (`Clamped` and `Clamped_RBV` both read 0, not the datatype default arriving late), and the clamped return drives readback and setpoint together, so putting 150 leaves both at 100 rather than the setpoint sitting at 150. Locally:

```
after connect:  readback=0   setpoint=0
after set(150): readback=100 setpoint=100
after set(42):  readback=42  setpoint=42
```

`test_ioc`'s child PVI assertion is updated for the new `Child:N:Clamped` PV. Note I could not run the p4p tests in my sandbox (`RuntimeError: Address family not supported by protocol` — no PVA-capable socket family), so that assertion update is verified by construction rather than by running it; CI is the real check. Say the word if you would rather have a plain non-clamping setter here.

---

tests/test_attributes.py
coretl: Done in ec2f44c4 — now `pytest.raises(RuntimeError, match="has no getter")`.

I also applied the same treatment to the four other bare `pytest.raises` in this file (the two `set_name`/`set_path` re-registration checks and the two `wait_for_*` timeouts), so the file is consistent rather than half-converted. Happy to narrow that back to just the two you flagged if you would rather keep the diff tight.

---

tests/test_attributes.py
coretl: Done in ec2f44c4 — now `pytest.raises(ValueError, match="do_update failed")`, which also pins that the getter's own exception is the one propagating rather than some other `ValueError` raised on the way through `update()`/`validate()`.

---

coretl:
Thanks @shihab-dls — the five inline comments are all addressed in ec2f44c4 (replies on each thread).

On the two runtime findings in your review body: I believe you were running a checkout from before the `Polled` rework, and both behave as you want on the current head. The giveaway is `poll_period=` — it stopped being a constructor argument in ab60286e (3 Aug), replaced by `Polled`/`NotPolled`, so on the head your second snippet raises before it gets anywhere:

```
TypeError: Attribute.__init__() got an unexpected keyword argument 'poll_period'
```

**1. Bare getter defaulting to `ONCE`.** No `TypeError` on the head — it constructs, reports `ONCE`, and schedules an initial poll:

```python
self.a = AttrRW(Int(), getter=self.getter)   # bare getter IS the ONCE schedule
```
```
constructed ok, poll_period = inf
api built; initial coros: 1 periodic: 0
after initial poll: readback = 10 setpoint = 10
```

I could not reproduce a `'<=' not supported between instances of 'object' and 'int'` anywhere on the head — there is no `object()` sentinel left in this path (`ONCE` is `float("inf")`), and the only `<=` against a period is the `@scan` decorator's positive-period check, which attributes do not go through.

**2. Setpoint seeding.** Your exact scenario, spelled for the head (getter returns 10, setter returns `value + 1`):

```
before connect:      readback=0  setpoint=0
after initial poll:  readback=10 setpoint=10    (you saw A_RBV=10 A=0)
after set(12):       readback=13 setpoint=13    (you saw A_RBV=13 A=13)
after set(15):       readback=16 setpoint=16    (you saw A_RBV=16 A=15)
```

This is the behaviour you asked for: the first *update* seeds the setpoint, not the first put. Transport-side seeding is gone entirely — `AttrRW.update` mirrors readback into the setpoint while `_setpoint_known` is `False`, and transports just register `add_setpoint_callback` (ADR 0020). The ordering holds at the IOC too: `serve()` calls `transport.connect()` (which registers the callbacks) before `_run_initial_coros()`, so the seeded setpoint reaches the records rather than being posted before anything is listening.

The `set(15)` → `A_RBV=16 A=15` inconsistency is gone for the same reason: a setter's bare return is the device's accepted value, so it drives both.

Since I am inferring the stale-checkout explanation rather than knowing it, could you re-pull and re-test at ec2f44c4? If either still misbehaves for you there I would very much like the traceback, because it would mean I am wrong about the cause.

---

shihab-dls:
> Thanks @shihab-dls — the five inline comments are all addressed in [ec2f44c](https://github.com/DiamondLightSource/fastcs/commit/ec2f44c4f24f4819f342b65cd24268e332bdf05b) (replies on each thread).
> 
> On the two runtime findings in your review body: I believe you were running a checkout from before the `Polled` rework, and both behave as you want on the current head. The giveaway is `poll_period=` — it stopped being a constructor argument in [ab60286](https://github.com/DiamondLightSource/fastcs/commit/ab60286e6b0b2f3b153a08e453eb6caf209bfe44) (3 Aug), replaced by `Polled`/`NotPolled`, so on the head your second snippet raises before it gets anywhere:
> 
> ```
> TypeError: Attribute.__init__() got an unexpected keyword argument 'poll_period'
> ```
> 
> **1. Bare getter defaulting to `ONCE`.** No `TypeError` on the head — it constructs, reports `ONCE`, and schedules an initial poll:
> 
> ```python
> self.a = AttrRW(Int(), getter=self.getter)   # bare getter IS the ONCE schedule
> ```
> 
> ```
> constructed ok, poll_period = inf
> api built; initial coros: 1 periodic: 0
> after initial poll: readback = 10 setpoint = 10
> ```
> 
> I could not reproduce a `'<=' not supported between instances of 'object' and 'int'` anywhere on the head — there is no `object()` sentinel left in this path (`ONCE` is `float("inf")`), and the only `<=` against a period is the `@scan` decorator's positive-period check, which attributes do not go through.
> 
> **2. Setpoint seeding.** Your exact scenario, spelled for the head (getter returns 10, setter returns `value + 1`):
> 
> ```
> before connect:      readback=0  setpoint=0
> after initial poll:  readback=10 setpoint=10    (you saw A_RBV=10 A=0)
> after set(12):       readback=13 setpoint=13    (you saw A_RBV=13 A=13)
> after set(15):       readback=16 setpoint=16    (you saw A_RBV=16 A=15)
> ```
> 
> This is the behaviour you asked for: the first _update_ seeds the setpoint, not the first put. Transport-side seeding is gone entirely — `AttrRW.update` mirrors readback into the setpoint while `_setpoint_known` is `False`, and transports just register `add_setpoint_callback` (ADR 0020). The ordering holds at the IOC too: `serve()` calls `transport.connect()` (which registers the callbacks) before `_run_initial_coros()`, so the seeded setpoint reaches the records rather than being posted before anything is listening.
> 
> The `set(15)` → `A_RBV=16 A=15` inconsistency is gone for the same reason: a setter's bare return is the device's accepted value, so it drives both.
> 
> Since I am inferring the stale-checkout explanation rather than knowing it, could you re-pull and re-test at [ec2f44c](https://github.com/DiamondLightSource/fastcs/commit/ec2f44c4f24f4819f342b65cd24268e332bdf05b)? If either still misbehaves for you there I would very much like the traceback, because it would mean I am wrong about the cause.
> 
> _Generated by [Claude Code](https://claude.ai/code)_

The first bug is fixed in the current head. However, the second bug is still present, but not for the reason I initially described. It seems that an `AttrRW` will sync it's setpoint with it's first readback value at startup, but this is not reflected on the PV. So, checking the actual setpoint value on the attribute will return the synced value, but doing a pvget/caget/etc on the setpoint PV will return 0. This is because, during the first update that is called via `_run_inital_coros()`, the PV is yet to be created, so the attributes `setpoint_callbacks` are empty. The callback to set the transport value is added to the attribute during `transport.serve()` which is in a `coros` list that is gathered after `_run_initial_coros()` is called, so the order results in the synced setpoint not being reflected on the transport.

tests/test_attributes.py
```diff
@@ -42,39 +39,121 @@ def test_attr_r():
...
+
+def test_datatype_required_when_not_inferable():
+    with pytest.raises(ValueError, match="datatype must be given explicitly"):
+        AttrR()
+
+    with pytest.raises(ValueError, match="datatype must be given explicitly"):
+        AttrW()
+
+    with pytest.raises(ValueError, match="datatype must be given explicitly"):
+        AttrRW()
```
shihab-dls: Should: we shouldn't input the same string multiple times like this, instead we should make a variable with the expected message at the start of the test, then match against the variable.

tests/test_attributes.py
```diff
@@ -153,50 +230,74 @@ async def device_add():
...
+@pytest.mark.asyncio
+async def test_set_setter_exception_is_caught_and_logged():
+    async def do_set(value):
+        raise ValueError("do_set failed")
+
+    attr = AttrW(Int(), setter=do_set)
+
+    # exception is caught and logged, not raised
+    await attr.set(5)
+    assert attr.setpoint == 5
```
shihab-dls, line 303: Should: this test only implies that the exception is caught and logged, but doesn't test that behaviour. One way to do this would be to also mock out the logger for this test, and check that it is called with the exception we expect.

tests/test_attributes.py
coretl: Done in 9c1ad3f6 — the message is hoisted to `expected_message` at the top of the test and all three `pytest.raises` match against it.

---

tests/test_attributes.py
coretl: Agreed — the test only proved the exception did not escape. Done in 9c1ad3f6, mocking the logger as you suggested:

```python
mock_logger = mocker.patch("fastcs.attributes.attr_w.logger")

await attr.set(5)
assert attr.setpoint == 5

logged_exception = mock_logger.opt.call_args.kwargs["exception"]
assert isinstance(logged_exception, ValueError)
assert str(logged_exception) == "do_set failed"
mock_logger.opt.return_value.error.assert_called_once_with(
    "Set failed", attribute=attr, setpoint=5
)
```

Asserting on `logger.opt(exception=...)` rather than just the message pins that it is the *setter's* `ValueError` being logged, not some other exception raised on the way through `set()`.

---

coretl:
You were right, and your second diagnosis was exactly right — thank you for chasing it down to the ordering. Fixed in 9c1ad3f6.

To confirm your analysis: it is **PVA-specific**. `EpicsCAIOC` builds its records in `__init__`, which runs during `transport.connect()` — before `_run_initial_coros()` — so CA already had its setpoint callback registered in time. `P4PIOC.__init__` only stored the APIs, and the PVs were built by `parse_attributes` inside `P4PIOC.run()`, i.e. inside `transport.serve()`, which is appended to `coros` and only awaited by the `asyncio.gather` *after* the initial polls have run. So `make_shared_write_pv` — and with it `add_setpoint_callback` — did not exist yet when the seed happened, and the seeded value went to an empty callback list. `attribute.setpoint` was correct; the PV served the datatype default.

This is why my earlier check did not catch it: I was asserting on `attribute.setpoint`, which is seeded correctly either way.

**The fix** is to make PVA follow the same contract CA already does — `connect()` builds the PVs, `serve()` runs the server:

```python
def __init__(self, controller_apis: list[ControllerAPI]):
    self._controller_apis = controller_apis
    self._providers = [parse_attributes(api) for api in self._controller_apis]

async def run(self):
    endless_event = asyncio.Event()
    with Server(self._providers):
        await endless_event.wait()
```

`parse_attributes` had no `await` in it, so it becomes a plain function rather than an `async def` that never yielded.

**Regression test** — `test_setpoint_seeded_by_initial_poll_reaches_transport` in `tests/transports/epics/pva/test_p4p.py`. It wraps the attribute's setpoint callback registration to record what gets published, calls `transport.connect()`, then runs the initial coros *without* ever awaiting `serve()` — mirroring the ordering in `FastCS.serve`:

```python
assert attribute.setpoint == 10
assert published == [10]
```

I checked it fails on the parent commit (`published == []`, i.e. your bug) and passes with the fix. It deliberately needs no p4p `Server`, so unlike the rest of `test_p4p.py` it runs in a sandbox without a PVA-capable socket family — and it would have caught this originally.

I also verified CA directly rather than assuming: capturing the out-record and running the initial poll gives `0` before the poll and `10` after, so `record.set()` before `iocInit()` does hold, and CA needs no change.

Verified with `uv run --locked tox -e pre-commit,type-checking` — both green. For `tests`, `pytest src tests --ignore=tests/benchmarking` gives 334 passed / 10 failed, and I confirmed the same 10 fail identically on the parent commit — all `RuntimeError: Address family not supported by protocol`, the sandbox's missing PVA socket family. Real CI is the check for those.

The two test comments from your latest review are addressed in the same commit, with replies on each thread.

---

